PR #5, opened 2020-06-12
meta: add license from nixos/nixpkgs

ghuntley:
It would be lovely to inline your work into nixpkgs itself. This PR is the first step to making it happen.


Trundle:
Thank you for your interest in NixOS-WSL! I'm not super familiar with nixpkgs - does it have to be this specific license or is it rather: "should have any FOSS license?" I'm asking because personally I would be way more comfortable to release this under Apache 2.0.

Trundle:
I went with Apache 2. I'm happy to relicense if that makes it easier that i will eventually end up in nixpkgs.

PR #9, opened 2021-01-09
Use Nix Flakes

JamieMagee:
Convert to using Nix Flakes and add GitHub Actions to automatically build a tarball﻿


Trundle:
Thank you, this is pretty awesome! I will have a look in the next days.

syschdemd.nix
```diff
@@ -1,19 +1,13 @@
...
 pkgs.substituteAll {
   name = "syschdemd";
   src = ./syschdemd.sh;
   dir = "bin";
   isExecutable = true;
 
-  buildInputs = [ daemonize ];
+  buildInputs = with pkgs; [ daemonize ];
 
-  inherit daemonize;
```
Trundle, line 16: This line is still required (respectively needs to be changed to `inherit (pkgs) daemonize;`), otherwise `@daemonize@` is not substituted in the result:

```
>  wsl -d NixOS
/nix/store/723ndl2p1p7rlm9dbiplglk04svbz82x-syschdemd/bin/syschdemd: line 19: @daemonize@/bin/daemonize: No such file or directory
```

JamieMagee:
@Trundle I think this is ready for merge now

One suggestion though: should the default username of `andy` be updated, to something like `nixos`, if this can now be provided as a prebuilt tarball?

Trundle:
Thank you! I changed the default username to `nixos` in 1c13caed8ac906de6de8e7726e674cd72316e7e3.

PR #12, opened 2021-02-18
Include the system configuration in the tarball

veehaitch:
Also include the system configuration in the tarball which was used to create it.


Trundle:
Thank you :pray:

PR #14, opened 2021-02-19
Support building without flake.nix

veehaitch:
Also support building without Nix Flakes (again) using

```shell
nix-build -A system -A config.system.build.tarball ./nixos.nix
```


Trundle:
I guess that counts as serial contributor now. Thank you :pray:

PR #15, opened 2021-02-20
Check format in workflow

veehaitch:
- Add a check to `flake.nix` which asserts that Nix source files a properly formatted using `nixpkgs-fmt`. 
- Run the check as part of the workflow
- Also execute the workflow for PRs


Trundle:
Thanks!

PR #17, opened 2021-03-06
Add workflow to update flakes

JamieMagee:
With this workflow a PR will be created weekly, on Sundays, to update the flake lockfile. The workflow can also be manually run


Trundle:
Thank you :pray: I'll have a closer look in the next days.

PR #19, opened 2021-03-16
Expose internal modules as flake nixosModules.

gcoakes:


Hoverbear (approved):
I've tested using this module in my own flake (https://github.com/Hoverbear-Consulting/flake/tree/wsl) and it works fine. Hope you don't mind if I keep using it! 😉 

I was originally just trying to package syschdemd (https://github.com/Hoverbear/NixOS-WSL/tree/packages.syschdemd) but your approach is much better.

Trundle:
The issue with this PR is that it breaks the non-flakes case: you can't `nixos-rebuild` from inside WSL.

gcoakes:
I stopped using WSL for a bit, but I decided to jump back into it a bit ago. I forgot this PR was even still up, and I started from the latest master a few days ago. The latest state of the branch should be fully backwards compatible even with the non-flake use case. @Trundle, I know you stepped back a bit from maintaining this project, but would you be able to comment whether the current iteration of this alleviates your previous concerns?

To give some background on my specific motivation for the changes, I use flakes to manage multiple system configurations with some shared modules (like my dev environment). I just want to consume the unique parts of NixOS-WSL like syschdemd and system.build.tarball, and I want to include the wsl module without conditional imports. So, these changes move most of the functionality behind enable options like `wsl.enable` and `wsl.tarball.enable`. I also added options to disable copying the config to `/etc/nixos` and setting the channels in the tarball. I keep my config in my home directory and just run `sudo nixos-rebuild switch --flake .#work`, so I don't want the default configs copied. Also, I don't use channels, so I don't want that copied either.

This would be breaking to the existing behavior except that all the defaults replicate the existing behavior faithfully. So, even people who consumed NixOS-WSL downstream by directly importing files shouldn't be impacted. `configuration.nix` imports `module.nix` and sets `wsl.enable = true;`. `build-tarball.nix` has all it's options set to true by default meaning that importers will see it almost as if it were written as `{ config.system.build.tarball = ... }` rather than with options.

Consuming the flake downstream is as simple as adding `NixOS-WSL.nixosModule` to the list of modules and setting `wsl.enable = true;`.

nzbr:
I have made similar changes to this in #56. It uses flake-compat in order to have everything (including the default system config) inside the flake. Having the option to disable copying the config to the tarball sounds great to me

gcoakes:
@nzbr, it looks like your PR has significantly more changes fixing a few issues I haven't even encountered yet. I'm more than happy to drop this PR if yours gets merged as it seems to accomplish the same goals.

I was very careful to keep my change backwards compatible even with the names of the files. This was in case someone downstream was consuming it by just importing build-tarball.nix or something similar. I did this so it was more palatable from a maintenance perspective. I would hate to delay upstreaming modularization because I did too many changes in one PR.

nzbr:
I have moved your idea to a new issue (#61), so I am closing this PR. Thank you though!

PR #25, opened 2021-04-10
Set locale when running nix activate

TeofilC:
This fixes the warnings given by perl each time a wsl container boots up


Trundle:
That's nice, it makes the first encounter with NixOS-WSL so much less threatening. Thank you!

TeofilC:
No worries! But this will only solve that on subsequent runs, though the PR (https://github.com/Trundle/NixOS-WSL/pull/26) I just opened should solve that rough edge as well

PR #26, opened 2021-04-10
Avoid having to run nix activate on first run

TeofilC:
This sets the initial /bin/sh to syschdemd, which means that the user no longer has to manually run nix activate.
Instead the user will be logged in to the default user account.

I've tested this with the `wsl.exe --import` install method but I haven't tested if it works with DistroLauncher.


Trundle:
Oh wow. That's clever. This makes the initial setup so much more pleasant. Thank you!

Issue #36, opened 2021-07-03
How to update?


anderspapitto:
I installed from a release tarball some time ago, and now I see that improvements have been made. How do I update my existing container to the current version of NixOS-WSL?

In particular, if I figure out a fix to #35 , I don't know how to test or apply my fix.

Congee:
The prebuilt one for now https://github.com/Trundle/NixOS-WSL/releases/tag/20.09-0cfd08f4881 is outdated as some patches have not been baked in. Installing a custom build tarball results in the notorious "Unspecified error".

So, I have run out of way of upgrading. :confused:

nzbr:
With the new releases, I have provided instructions on how to update from previous versions. If you have a setup that is not covered in the release notes and need assistance with updating, please open a discussion in the "Q&A" category

PR #37, opened 2021-07-03
start a login shell at login

### LINKED ISSUE #35: Initial shell is not a login shell, does not source ~/.bash_profile
I'm seeing that `~/.bash_profile` is not being sourced automatically when opening a wsl terminal, but it is if I then run `bash --login`. In the initial shell, I've also checked with this command `shopt -q login_shell && echo 'Login shell' || echo 'Not login shell'`, which I got from https://unix.stackexchange.com/a/26782.

In particular, I'm trying to use home-manager inside wsl, and the `programs.bash.sessionVariables` option doesn't work because of this.

anderspapitto:
fixes #35

unsure how to test or apply, see #36

### DIFF
--- a/syschdemd.sh
+++ b/syschdemd.sh
@@ -38,6 +38,6 @@ if [[ $# -gt 0 ]]; then
     shift
     cmd="$@"
 else
-    cmd="$userShell"
+    cmd="$userShell --login"
 fi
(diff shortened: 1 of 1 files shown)


syschdemd.sh
```diff
@@ -38,6 +38,6 @@ if [[ $# -gt 0 ]]; then
     shift
     cmd="$@"
 else
-    cmd="$userShell"
+    cmd="$userShell --login"
```
ajgrf, line 41: The short flag `-l` is more portable than `--login`, which doesn't work for dash or mksh, for example.

anderspapitto:
closing this myself due to inactivity, though it may still be useful to others

PR #39, opened 2021-07-22
Add support for WSLg

nzbr:
This PR adds support for [WSLg](https://github.com/microsoft/wslg).
GUI Applications can be launched from a shell as well as from the start menu

I added an activation script, that copies the application launchers to `/usr/share/applications` so that the installed applications will appear in the windows start menu

`syschdemd.sh` was changed to start `/bin/sh` as a login shell because the environment variables aren't present otherwise

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -48,17 +48,6 @@ in
   # Don't allow emergency mode, because we don't have a console.
   systemd.enableEmergencyMode = false;
 
-
-  # WSLg support
-  environment.variables = {
-    DISPLAY = ":0";
-    WAYLAND_DISPLAY = "wayland-0";
-
-    PULSE_SERVER = "${automountPath}/wslg/PulseServer";
-    XDG_RUNTIME_DIR = "${automountPath}/wslg/runtime-dir";
-    WSL_INTEROP = "/run/WSL/1_interop";
-  };
-
   environment.etc."wsl.conf".text = ''
     [automount]
     enabled=true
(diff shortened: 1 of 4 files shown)


nzbr:
The solution proposed in #40 sounds more future-proof than this one. The activation script would still be needed for the start menu application launchers to work though

nzbr:
This PR now includes the changes from #45 instead of hard-coding the environment variables. New changes included here are the customizable mount path for windows drives as well as copying application launchers to `/usr/share/applications` so that windows creates the appropriate start menu entries

configuration.nix
```diff
@@ -13,6 +14,9 @@ in
   # WSL is closer to a container than anything else
   boot.isContainer = true;
 
+  # Include Windows %PATH% in Linux $PATH.
+  environment.extraInit = ''PATH="$PATH:$WSLPATH"'';
```
SuperSandro2000, line 18: That has one downside: It slows down tab completion down tremendously because looking up the windows directories is very very very slow.

See https://github.com/akinomyoga/ble.sh/issues/96#issuecomment-912952486= for some details.

nzbr: From what I understand, this is a problem for all WSL distros. Appending the path can be toggled off with `wsl.wslConf.interop.appendWindowsPath = false;`. #64 has an option to toggle off the extraInit code

SuperSandro2000: Yes, it is sadly.

PR #41, opened 2021-08-10
Gather env vars from WSL and merge with NixOS.

Hajitorus:
I tried using https://github.com/Trundle/NixOS-WSL/pull/40 to bring in WSL variables, but using that meant I lost any environment from the NixOS side (especially NixOS's PATH).

I'm aware this is stupid and hacky, but I wanted to flag that just going to systemd-run broke my setup because PATH only had /usr/bin etc after applying above-mentioned patch.

With this patch, my WSL2 environment loads and WSL interop works.

### DIFF
--- a/syschdemd.sh
+++ b/syschdemd.sh
@@ -40,4 +40,23 @@ if [[ $# -gt 0 ]]; then
 else
     cmd="$userShell"
 fi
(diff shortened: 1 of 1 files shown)


Hajitorus:
#45 fixes this much more cleanly. Marking closed.

PR #40, opened 2021-07-26
Pass outer environment to user session

vlaci:
Changed machinectl to systemd-run to have finer grained control on how the user session is initiated. It has the following advantages:
* Direct support for specifying working directory
* Can be passed arbitrary directives understood by Systemd (see `systemd.exec(5)` for details). It is used to pass environment variables.
* Returns proper exit status of internal commands (machinectl swallows it)

By passing these variables WSLg and windows executables now just work out of the box. Hopefully any new environment variables introduced in the WSL ecosystem in the future will just work as well.

Fixes #32

Issue #29, opened 2021-04-24
Machinectl takes over tty


tudurom:
Hello, I'm trying to use GPG but it doesn't work because pinentry cannot open a tty.

```
$ echo GETPIN | pinentry -T /dev/pts/0
No $DBUS_SESSION_BUS_ADDRESS found, falling back to curses
OK Pleased to meet you
S ERROR gnome3.open_tty_for_read 83918849
ERR 83918849 Permission denied <Pinentry>
```

It works just fine if I start a new tty with `script /dev/null`:

```
$ script /dev/null
$ tty
/dev/pts/1
$ echo GETPIN | pinentry -T /dev/pts/1
```

Also there should be a way to have a systemd user session running, at least for dbus support. That way, graphical pinentry (GTK3) would work which is better.

Reilithion:
I had a related issue when trying to enable `programs.ssh.startAgent` (in that having a user session might help?). Basically, the ssh agent wasn't starting. So I set the default user's password with `sudo passwd` and ran `sudo login`. Once I logged in, the ssh agent was running. Can you see if this workaround also has an effect on your issue?

tudurom:
This worked very well! Thank you! The reason ssh agent works is because it seems that this also starts up systemd’s user session.

A curious thing I noticed is that even though before running `sudo login` the command `echo GETPIN | pinentry -T /dev/pts/0` would fail, causing gpg-agent to fail too, after running `sudo login` *and* closing the shell the command would also work, meaning that the tty was altered.

tudurom:
There is one downside to this `sudo login` hack: graphical applications don't work anymore.

tudurom:
This was fixed in 89c644e7178abc4abf51e4d1c82222ec18f9878a.

Issue #18, opened 2021-03-09
Systemd user process


Cjen1:
Hi, this seems like a great project. I've run into an issue where the systemd user manager doesn't appear to be started (meaning that systemd user processes can't run).

I'm trying to track down why this is occurring (may just be due to lazy evaluation tbh).

Specific problem:
```
nixos@Shale > systemctl status --user
Failed to connect to bus: No such file or directory
```

```
nixos@Shale /m/c/U/Cjen0 [255]> ps aux | grep systemd
root         1  0.1  0.0  22280 10676 ?        Ss   12:18   0:00 systemd
root        21  0.0  0.0  36996 11644 ?        Ss   12:18   0:00 /nix/store/q6ylicsava582g55mdx0788yg2fib7js-systemd-246.6/lib/systemd/systemd-journald
root        48  0.0  0.0  16568  5904 ?        Ss   12:18   0:00 /nix/store/q6ylicsava582g55mdx0788yg2fib7js-systemd-246.6/lib/systemd/systemd-logind
message+    50  0.0  0.0   6288  3504 ?        Ss   12:18   0:00 /nix/store/4013xmv3pjddrdgb6vkncb4iqlyij8pq-dbus-1.12.20/bin/dbus-daemon --system --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only
nixos      365  0.0  0.0 223440  2260 pts/0    S+   12:19   0:00 /nix/store/3mx947xcmsyk79izqc8ifv31qafp80pc-gnugrep-3.6/bin/grep --color=auto systemd
```

While the equivalent run on a server:
```
cjen1@____:~$ ps aux | grep systemd
root         1  0.0  0.0 226592  7784 ?        Ss    2020  22:15 /lib/systemd/systemd --system --deserialize 23
root      1243  0.0  0.0  65628  5484 ?        Ss    2020   3:19 /lib/systemd/systemd-logind
message+  1261  0.0  0.0 154432  3644 ?        Ss    2020  10:09 /usr/bin/dbus-daemon --system --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only
cjen1     4893  0.0  0.0  14436  1072 pts/0    S+   12:20   0:00 grep systemd
cjen1    10568  0.0  0.0  39476  3748 ?        Ss   Feb16   0:00 /usr/bin/dbus-daemon --session --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only
[...]

PR #45, opened 2021-10-13
Yet another environment pass-through PR

ajgrf:
- Compared to #39: I did not hard-code anything, so there won't be an ever-growing list of environment variables we need to manually pass through.
- Compared to #40: I did not switch to `systemd-run`, therefore systemd still starts user sessions. Avoids reintroducing issues like #18 and #29.
- Compared to #41: My patch is simpler and doesn't combine `machinectl` and `systemd-run` in confusing ways.

I also added the Windows %PATH% to the Linux $PATH, so running Windows exe's doesn't require specifying the full absolute file path.

### DIFF
--- a/syschdemd.sh
+++ b/syschdemd.sh
@@ -40,4 +40,10 @@ if [[ $# -gt 0 ]]; then
 else
     cmd="$userShell"
 fi
-exec $sw/nsenter -t $(< /run/systemd.pid) -p -m -- $sw/machinectl -q --uid=@defaultUser@ shell .host /bin/sh -c "cd \"$PWD\"; exec $cmd"
+
+# Pass external environment but filter variables specific to root user.
+exportCmd="$(export -p | $sw/grep -vE ' (HOME|LOGNAME|SHELL|USER)=')"
+
+exec $sw/nsenter -t $(< /run/systemd.pid) -p -m -- $sw/machinectl -q \
+	--uid=@defaultUser@ shell .host /bin/sh -c \
+	"cd \"$PWD\"; $exportCmd; exec $cmd"
--- a/syschdemd.sh
+++ b/syschdemd.sh
@@ -42,7 +42,7 @@ else
 fi
 
 # Pass external environment but filter variables specific to root user.
-exportCmd="$(export -p | $sw/grep -vE ' (HOME|LOGNAME|SHELL|USER)=')"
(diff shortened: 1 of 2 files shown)


ajgrf:
Did you shutdown the WSL distro after building from my branch (i.e. run `wsl -t NixOS`)? It needs a fresh start to take effect.

K900:
I've been running this for a while now and it works perfectly.

felipeqq2:
Just rebuilt the tarball and it's working just fine. I probably got something wrong somewhere 😅

syschdemd.sh
```diff
@@ -40,4 +40,10 @@ if [[ $# -gt 0 ]]; then
...
     cmd="$userShell"
 fi
-exec $sw/nsenter -t $(< /run/systemd.pid) -p -m -- $sw/machinectl -q --uid=@defaultUser@ shell .host /bin/sh -c "cd \"$PWD\"; exec $cmd"
+
+# Pass external environment but filter variables specific to root user.
+exportCmd="$(export -p | $sw/grep -vE ' (HOME|LOGNAME|SHELL|USER)='); export WSLPATH=\"$PATH\""
+
+exec $sw/nsenter -t $(< /run/systemd.pid) -p -m -- $sw/machinectl -q \
+	--uid=@defaultUser@ shell .host /bin/sh -c \
+	"cd \"$PWD\"; $exportCmd; exec $cmd"
```
puffnfresh, line 49:
```suggestion
exec $sw/nsenter -t $(< /run/systemd.pid) -p -m --wd="$PWD" -- $sw/env WSLPATH="$PATH" @wrapperDir@/su -s $userShell @defaultUser@ "$@"
```

I'm just passing through `PATH` as `WSLPATH` and using your change to `configuration.nix`. Works really well.

I'm not sure why you have the rest, I think other environment variables are already passed through:

```
$ env | grep WSL
WSL_INTEROP=/run/WSL/807_interop
WSL_DISTRO_NAME=NixOS
WSLENV=WT_SESSION::WT_PROFILE_ID
```

So a simple `env` seems like enough.

ajgrf: `machinectl` is needed to properly start a user session, and it doesn't pass through the environment on its own. By replacing the call to `machinectl`, `systemd` won't start user services, for example.

PR #53, opened 2022-02-01
Fix root shell (Fixes #51)

### LINKED ISSUE #51: sudo -i fails with `nsenter: cannot open /proc/136/ns/pid: No such file or directory`
Right after the install:

```
$ sudo -i
nsenter: cannot open /proc/136/ns/pid: No such file or directory
```

This is on `20.09-0cfd08f4881` (which is the latest and the only release)

nzbr:
When a root shell is started from within the systemd namespace, syschdemd.sh will try to enter the namespace again which causes the command to fail. This PR adds a new environment variable INSIDE_NAMESPACE that is used to detect this and start the requested command directly

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -29,7 +29,12 @@ in
     extraGroups = [ "root" ];
   };
 
-  security.sudo.wheelNeedsPassword = false;
+  security.sudo = {
+    extraConfig = ''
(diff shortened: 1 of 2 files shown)


G3zz (commented):
This is great and fixed my problem - thanks!

nzbr:
The fix was a part of #56

PR #56, opened 2022-02-15
Everything that changed in my fork of this repo

### LINKED ISSUE #30: unable to access root super user:
Hello @Trundle ,
I'm unable to access root in nixos-wsl using su? I'm just wondering if that's normal, and not an issue?
Anyways, I've been able to get by with running `sudo commands ...`, which works, but this way does not seem secure.

If you could clarify this, thanks in advance for the help!

### LINKED ISSUE #32: Does running .exe's still work with the custom init system?
Apologies in advance for posting the question in the title as an issue (didn't notice anywhere I could post questions regarding this repo), though I'm looking to migrate from NixOS to Windows w/ WSL2 + NixOS and wanted to know if this is doable.

The use case is for writing Windows code in WSL2, and debugging them directly on Windows. As of now w/ NixOS, I use Wine to debug Windows code I write.

### LINKED ISSUE #34: Mount windows drives with metadata option
Hi, first of all: Awesome work! :)

To use git and some other programs from inside a WSL distro and access the filesystem "outside", which is (probably in most cases?) an NTFS formated drive, one needs to instruct WSL to mount the Windows drives with the metadata option.

This is done via `/etc/wsl.conf` (*inside* the disto):
~~~
[automount]
options = "metadata"
~~~

Therefore I've added
~~~
  environment.etc."wsl.conf" = {
    text = ''
      [automount]
      options = "metadata"
           '';
  };
~~~
to `/etc/nixos/configuration.nix`.

I figured this might be useful for some.
Fixes issues such as `fatal: could not set 'core.filemode' to 'false'` messages when using `git clone` or `git init`.

Reference: https://askubuntu.com/a/1242671/635446

### LINKED ISSUE #35: Initial shell is not a login shell, does not source ~/.bash_profile

### LINKED ISSUE #50: Docker using the host
Is it possible to use Docker from docker desktop like with other distros? It's basically the only thing that makes me keep using ubuntu as a escape hatch.

I currently get this output when trying:
```
> docker ps
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/json": dial unix /var/run/docker.sock: connect: permission denied
```

### LINKED ISSUE #51: sudo -i fails with `nsenter: cannot open /proc/136/ns/pid: No such file or directory`

### LINKED ISSUE #57: [Windows 11] systemctl --user fails with release 20.09-0cfd08f4881
On Windows 11
```
[nixos@nixos:~]$ systemctl --user status
Failed to connect to bus: No such file or directory
```

strace shows `connect(3, {sa_family=AF_UNIX, sun_path="/mnt/wslg/runtime-dir/systemd/private"}, 40) = -1 ENOENT (No such file or directory)`. This pointed me to Windows 11/wslg.

Replacing `/etc/nixos/syschdemd.sh` with https://raw.githubusercontent.com/nix-community/NixOS-WSL/main/syschdemd.sh fixes `systemctl --user`.

However, now vscode-server breaks. Going with https://raw.githubusercontent.com/nzbr/NixOS-WSL/main/syschdemd.sh fixes also this part.

nzbr:
The changes that I have implemented in my fork of this repo can't be separated easily anymore, so I have put all of them in one PR.

#53, #45 and #39 are integrated in this pull request

- The configuration has been split into multiple modules, which are exposed through the flake
	- the system configuration accesses the flake using flake-compat, so that it can be built without a flake-enabled nix
- Optional Docker Desktop support (can be enabled in configuration.nix)
- nixos.nix has been removed in favor of using flake-compat for nix-build. The CI pipeline has been adjusted accordingly
- syschdemd.sh loads /etc/set-environment before starting a command. This makes  the WSL-Remote in VSCode work if https://github.com/msteen/nixos-vscode-server is used
- A module has been added to build an installer tarball.
    - This wraps around the system tarball and should be used with `wsl --import` instead. Importing the system tarball can fail with an "Unspecified Error". One of the causes for this error (at least in my testing) is any hard-link being present in the tarball. The installer mitigates this issue by including only a statically-compiled version of busybox, that is used to unpack the system tarball when starting the system for the first time

Fixes #30
Fixes #32
Fixes #34
Fixes #35
Fixes #50
Fixes #51
Fixes #57

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,7 @@
+use_flake() {
+    watch_file flake.nix
+    watch_file flake.lock
+    eval "$(nix print-dev-env)"
+}
+
+use flake
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,10 +28,21 @@ jobs:
         run: |
           nix build '.#nixosConfigurations.mysystem.config.system.build.tarball'
 
-      - uses: actions/upload-artifact@v2
+      - name: Upload tarball
+        uses: actions/upload-artifact@v2
         with:
-          name: install.tar.gz
+          name: rootfs
           path: result/tarball/nixos-wsl-x86_64-linux.tar.gz
+
+      - name: Build installer
+        run: |
+          nix build '.#nixosConfigurations.mysystem.config.system.build.installer'
+
+      - name: Upload installer
+        uses: actions/upload-artifact@v2
+        with:
+          name: installer
+          path: result/tarball/nixos-wsl-installer.tar.gz
 
   release:
     if: startsWith(github.ref, 'refs/tags/')
@@ -43,7 +54,11 @@ jobs:
 
       - uses: actions/download-artifact@v2
         with:
-          name: install.tar.gz
+          name: rootfs
+
+      - uses: actions/download-artifact@v2
+        with:
+          name: installer
 
       - name: Create Release
         id: create_release
@@ -59,7 +74,17 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          upload_url: ${{ steps.create_release.outputs.upload_url }} #
-          asset_name: install.tar.gz
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_name: nixos-wsl-x86_64-linux.tar.gz
           asset_path: result/tarball/nixos-wsl-x86_64-linux.tar.gz
+          asset_content_type: application/gzip
+
+      - name: Upload installer as release asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_name: nixos-wsl-installer.tar.gz
+          asset_path: result/tarball/nixos-wsl-installer.tar.gz
           asset_content_type: application/gzip
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,7 @@ jobs:
       - uses: actions/upload-artifact@v2
         with:
           name: install.tar.gz
-          path: result/tarball/nixos-system-x86_64-linux.tar.gz
+          path: result/tarball/*.tar.gz
 
   release:
     if: startsWith(github.ref, 'refs/tags/')
@@ -60,6 +60,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }} #
-          asset_name: nixos-system-x86_64-linux.tar.gz
-          asset_path: nixos-system-x86_64-linux.tar.gz
+          asset_name: install.tar.gz
+          asset_path: result/tarball/*.tar.gz
           asset_content_type: application/gzip
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,7 @@ jobs:
       - uses: actions/upload-artifact@v2
         with:
           name: install.tar.gz
-          path: result/tarball/*.tar.gz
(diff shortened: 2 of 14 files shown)


b3vis:
> I have actually used native docker in NixOS-WSL for quite a while. That is until the docker daemon did not start anymore (probably after a nixpkgs update). If you find a way to fix it, that'd be great!

It's because in 21.11 we switched from using iptables to nftables through a compatibility layer. You can set an extraOptions and do `--iptables=false` but then I found none of the containers have outbound network connectivity. I've got this working by sticking with 21.05 for now with WSL2. 

I was going to try and pin docker, systemd & iptables/nftables to 21.05 nixpkgs in my flake.nix but this feels janky and not the solution.

nzbr:
The interesting thing is, that I have multiple hosts (with NixOS on hardware) on 21.11, which run docker just fine. They are using the exact same revision of nixpkgs as my WSL host

b3vis:
> The interesting thing is, that I have multiple hosts (with NixOS on hardware) on 21.11, which run docker just fine. They are using the exact same revision of nixpkgs as my WSL host

Same, I’m using 21.11 in production with zero issues with docker. There’s something funky around it running in WSL.

elebeaup:
I'm using 21.11 in WSL and with docker but I had to recompile the WSL kernel by enabling all nftables modules

b3vis:
> I'm using 21.11 in WSL and with docker but I had to recompile the WSL kernel by enabling all nftables modules

How did you do this @elebeaup - Do you have some example nix for the modules needed?

elebeaup:
1. Clone the WSL kernel source
```shell
$ git clone --depth 1 https://github.com/microsoft/WSL2-Linux-Kernel.git
```

2. Enable the nftables modules
```shell
$ cat <<EOT >> Microsoft/config-wsl
CONFIG_NFT_CT=y
CONFIG_NFT_COUNTER=y
CONFIG_NFT_CONNLIMIT=y
CONFIG_NFT_LOG=y
CONFIG_NFT_LIMIT=y
CONFIG_NFT_MASQ=y
CONFIG_NFT_REDIR=y
CONFIG_NFT_NAT=y
CONFIG_NFT_TUNNEL=y
CONFIG_NFT_OBJREF=y
CONFIG_NFT_QUEUE=y
CONFIG_NFT_QUOTA=y
CONFIG_NFT_REJECT=y
CONFIG_NFT_REJECT_INET=y
CONFIG_NFT_COMPAT=y
CONFIG_NFT_HASH=y
CONFIG_NF_SOCKET_IPV4=y
CONFIG_NF_TPROXY_IPV4=y
CONFIG_NFT_REJECT_IPV4=y
CONFIG_NF_SOCKET_IPV6=y
CONFIG_NF_TPROXY_IPV6=y
CONFIG_NFT_REJECT_IPV6=y

# CONFIG_KCSAN is not set
EOT
```

3. Compile the kernel with nftables support
```shell
$ make KCONFIG_CONFIG=Microsoft/config-wsl
```

4. Copy the custom kernel image (i.e. arch/x86/boot/bzImage) to the windows filesystem

5. Configure the .wslconfig file as follow:
```
[wsl2]
kernel=C:\\temp\\myCustomKernel
```

nzbr:
Thank you @TeofilC! I used the workaround to imporove upon the installer. I don't really want to modify the final system's derivations by replacing hardlinks with symlinks there. It should't be a problem for the installer tarball though, because none of those files remain in the installed system anyway

TeofilC:
Nice! I'm glad you found it helpful

PR #65, opened 2022-03-28
NixOS 21.11

nzbr:
This seems to be working without any changes besides switching to the new channel. It'd be great if someone else could verify that everything works before I merge this. (Just in case I missed something)


K900:
Not sure about 21.11, but I've been running unstable for a while now and it works fine.

grantbevis:
I’ve been using 21.11 and it’s fine

PR #79, opened 2022-04-16
Fix Windows File Systems not being mounted on first start when using the Installer

### LINKED ISSUE #72: /bin/sh: line 1: cd: /root: Permission denied
I imported a fresh NixOS using:
```
wsl --import NixOS .\NixOS\ .\Downloads\nixos-wsl-installer-fixed.tar.gz --version 2
```

When I start it I get:
```
wsl -d NixOS
/bin/sh: line 1: cd: /root: Permission denied

[nixos@nixos:~]$
```
(Also `/mnt/*` is not mounted.)

After restarting the distro it seems to work again:
```
> wsl -t NixOS
> wsl -d NixOS
Copying /usr/share/icons
setting up /etc...

[nixos@myhostname:/mnt/c/Users/michas]$
```

nzbr:
Add a `/bin/mount` binary to the installer, so that WSL can correctly mount the file systems
`/etc/fstab` is not required, but an error appears if it is not present

When the shell is started in `/root`, syschdemd now silently changes to the default user's home directory, in order to prevent an error message from appearing. This happens, when WSL is started in a directory, that is not available in the guest fs

Closes #72

### DIFF
--- a/modules/installer.nix
+++ b/modules/installer.nix
@@ -54,7 +54,10 @@ with builtins; with lib; {
 
         contents = [
           { source = config.environment.etc."wsl.conf".source; target = "/etc/wsl.conf"; }
+          { source = config.environment.etc."fstab".source; target = "/etc/fstab"; }
           { source = passwd; target = "/etc/passwd"; }
+          { source = "${pkgs.busybox}/bin/busybox"; target = "/bin/sh"; }
+          { source = "${pkgs.busybox}/bin/busybox"; target = "/bin/mount"; }
         ];
       };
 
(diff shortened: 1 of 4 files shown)


sielicki:
I was able to consistently reproduce this when WSL was started with Windows Terminal Preview, can confirm that the first commit ( [fix "cd: /root: Permission denied"](https://github.com/nix-community/NixOS-WSL/pull/79/commits/67b9d7c14322667e2a30405edcc812164b73738f)[](https://github.com/nzbr) ) is good.

PR #83, opened 2022-04-21
WIP: Windows Hello Authentication

nzbr:
This adds an option to approve sudo and other login prompts through Windows Hello (Fingerprint/Face/PIN/etc.) instead of a password or not at all (as is the default)

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -21,6 +21,7 @@ in
     # docker.enable = true;
 
     # Enable authenticating sudo prompts with Windows Hello
+    # DO NOT USE THIS FOR ANYTHING SECURITY-CRITICAL
     # windowsHello.enable = true;
   };
 


nzbr:
This seems to stop working sometimes (exit code 255 from the .exe). That can be fixed by opening a new shell - it's probably related to the WSL_INTEROP socket

modules/windows-hello.nix
```diff
@@ -0,0 +1,50 @@
...
+                dontUnpack = true;
+                buildCommand = ''
+                  install -m 0755 $src $out
+                '';
+              };
+              wrapper = pkgs.writeShellScript "wrapper" ''
+                export PATH=${pkgs.coreutils}/bin # The PAM environment does not include the default PATH
+                export WSL_INTEROP="/run/WSL/$(ls -tr /run/WSL | tail -n1)" # Find the correct WSL_INTEROP socket to be able to start the EXE
+                exec ${authenticator} [$PAM_SERVICE] $PAM_RUSER wants to authenticate as $PAM_USER
+              '';
```
SuperSandro2000, line 42: I think we should make a proper package out of that. Could we build it from source, too?

SuperSandro2000:
Also https://github.com/nzbr/PAM-WindowsHello#--%EF%B8%8F-you-probably-do-not-want-to-use-this-for-anything-security-critical-%EF%B8%8F---i-am-not-responsible-for-anyone-bypassing-the-login-on-a-system-you-install-this-to

modules/windows-hello.nix
nzbr: I don't think it'd be possible to build from source. The program has build-time dependencies against `Windows.winmd` and `System.Runtime.WindowsRuntime.dll`
If there's any way to get those files into the nix store, it might be possible to build with Mono though

SuperSandro2000: Both are on nuget https://www.nuget.org/packages/Microsoft.Windows.WinMD/ https://www.nuget.org/packages/System.Runtime.WindowsRuntime/

nzbr: I didn't expect that, thanks! I'll look into it

nzbr: I tried it now (the correct package is https://www.nuget.org/packages/Microsoft.Windows.SDK.Contracts btw. The one above seems to be for parsing WinMD files), however I could not get msbuild on Linux to compile the executable. It always threw an error about a missing assembly reference. I also tried porting the helper to .NET 6, but that wouldn't build on Linux either because `NETSDK1100: Windows is required to build Windows desktop applications`. I had initially meant to build it in .NET 6, but couldn't figure out how to import WinRT then. In retrospect that was fortunate, because now I learned that .NET 6 would require having an additional .dll in the same directory as the .exe, whereas .NET Framework only requires the .exe to be present

TL;DR Unfortunately it looks like there is no way to build the helper on linux

aikooo7:
What needs to be done to this be merged?

nzbr:
The Nix code in this PR is pretty outdated and would probably need to be adjusted to the current state of the main branch. Also the windows binary this uses can't be cross-compiled on linux and therefore needs to be pulled in as a binary blob. If that's possible the windows part should be rewritten in something like rust where that's possible, so that it can be built in a derivation

Cu3PO42:
I actually did a rewrite of both the PAM module and the Windows side (in Rust) for this roughly two years ago. It also doesn't use any unsupported APIs anymore and should cross-compile. It worked, but I have a feeling it wasn't fully done. It's currently rotting in a private repo, but I'll see about cleaning it up.

nzbr:
That'd be really cool, tbh

maicol07:
> I actually did a rewrite of both the PAM module and the Windows side (in Rust) for this roughly two years ago. It also doesn't use any unsupported APIs anymore and should cross-compile. It worked, but I have a feeling it wasn't fully done. It's currently rotting in a private repo, but I'll see about cleaning it up.

@Cu3PO42 any news about this?

Thanks

PR #90, opened 2022-05-01
fix docker-desktop systemd proxy

### LINKED ISSUE #89: Docker Desktop WSL2 not working
## Bug description
When I enable the option `docker.enable = true;` in my `configuration.nix` I expect then to see my running docker images in Docker Desktop for Windows when executing `docker ps`. Instead I get:

```bash
$ docker ps
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

So my first thought here is, why is it trying to connect to the local docker socket and not the proxy?

After some trials and errors I finally found following hint:

```bash
/mnt/wsl/docker-desktop/docker-desktop-proxy -docker-desktop-root /mnt/wsl/docker-desktop
bash: /mnt/wsl/docker-desktop/docker-desktop-proxy: No such file or directory
```

On my system it seems the name is different, instead when running this command, everything works:

```bash
/mnt/wsl/docker-desktop/docker-desktop-user-distro proxy --docker-desktop-root /mnt/wsl/docker-desktop
```

Before submitting a Pull-Request with a fix that works only for me, I want to figure out how comes that my system behaves differently from the others. In the next step I will let you know the versions of NixOS-WSL and Docker Desktop I use, to let you potentially reproduce this issue.

## To Reproduce

- Host: Windows 11 Pro; Version: 21H2; OS build: 22000.613
- This is the tarball I used: https://github.com/nix-community/NixOS-WSL/releases/download/21.11-d89f18a1/nixos-wsl-installer-fixed.tar.gz
- Docker Desktop for Windows Version 4.7.1 (77678)
  - Settings > Resources > WSL Integration
  - [x] Enable Integration with my default WSL distro
  - [ ] NixOS
  - [x] Ubuntu
  - NixOS is **not** my default WSL distro

hschaeidt:
Potentially fixes #89

This is the changeset I did to fix docker-desktop on my machine. I think it needs further investigation of why I have different paths than other users. See #89 for more information.

### DIFF
--- a/modules/docker-desktop.nix
+++ b/modules/docker-desktop.nix
@@ -23,7 +23,7 @@ with builtins; with lib; {
       systemd.services.docker-desktop-proxy = {
         description = "Docker Desktop proxy";
         script = ''
-          ${config.wsl.automountPath}/wsl/docker-desktop/docker-desktop-proxy -docker-desktop-root ${config.wsl.automountPath}/wsl/docker-desktop
+          ${config.wsl.automountPath}/wsl/docker-desktop/docker-desktop-user-distro proxy --docker-desktop-root ${config.wsl.automountPath}/wsl/docker-desktop
         '';
         wantedBy = [ "multi-user.target" ];
         serviceConfig = {


nzbr:
Disabling mount isolation causes a lot of interesting new problems, so that's not an option. Another option might be to somehow start the proxy outside the systemd namespace

nzbr:
It works now. I'll open a second PR for that

Issue #63, opened 2022-03-24
Package for Windows Store


sielicki:
It would be nice to just have this available on the Windows Store. It's more or less just a matter of making a manifest and emailing the right people.

See: github.com/microsoft/WSL-DistroLauncher

sielicki:
I started looking through this tonight. I'd be interested to hear ideas around how we could take advantage of having an executable in the default Windows `%PATH%` to do more interesting nix-on-wsl things, especially with regards to switching and rolling back generations.

There's a nice little API within there where you can run arbitrary commands within WSL through `WslLaunchInteractive` to call whatever you want.

nzbr:
I have implemented a custom distro launcher (#94). [A test-signed build can be found here](https://github.com/nzbr/NixOS-WSL/suites/6740693791/artifacts/257347803). To install it, run the included `install.ps1` script in PowerShell. It may be necessary to enable Developer Mode in the Windows settings

If you discover any issues, please leave a comment on the Pull Request

nzbr:
The build artifact I linked has probably expired since then. You should alwasys be able to grab the `appx` artifact from the latest CI run here: https://github.com/nzbr/NixOS-WSL/actions?query=branch%3Adistro-launcher

luochen1990:
Has there been any progress?

SuperSandro2000:
Not to much on #94

FilipeBento:
Hi, is there any progress towards having a NixOS in the Microsoft Store to use in WSL2?

aikooo7:
> Hi, is there any progress towards having a NixOS in the Microsoft Store to use in WSL2?

The distrolauncher is already done, but I'm not sure how we can put it in the store and since @nzbr is the maintainer, maybe should be he doing it?

nzbr:
The latest CI runs already build the store package. It'd be great if some of you would test the thing to see if it works on other computers than mine. Just grab the `launcher-msix` [artifact from here](https://github.com/nix-community/NixOS-WSL/actions/runs/7866791929#artifacts), unpack it and run the `Install.ps1` inside. ([right-click] -> [run with powershell] should do the trick)

aisk:
Hi @nzbr after running the Install.ps1 succeeded, the NixOS item showed in the start menu, but it quite imediately after click. Run `nixos` in command line shows:

```
C:\Users\xxxxx>nixos
Registering NixOS...
There was an error registering the distribution
```

nzbr:
[...]

PR #94, opened 2022-05-22
DistroLauncher

nzbr:
Add a custom DistroLauncher. This will allow packaging NixOS-WSL as an appx package and upload it to the MS Store eventually

This is a custom reimplementation of DistroLauncher in C# (instead of C++, which is what the reference implementation uses)
The rewrite gives us greater flexibility in extending the DistroLauncher with additional features, e.g. upgrading to newer releases of NixOS-WSL with something like `nixos update`

- [x] Fix WslLaunchInteractive return codes
- [x] Updating works
- [x] systemd implementation in version output
- [x] Update.cs: Use mktemp instead of a hard-coded directory
- [ ] Check for nixos-unstable channel before overwriting it
- [ ] Make secrets more easily accessible by using something like age or sops
- [ ] Put a decryption key into the repo secrets

_Future ideas_
- [ ] Rescue Shell
- [ ] Rollback
- [ ] Updating to specific git revisions instead of the bundled tarball's version

Related: #63 #78


nzbr:
WslLaunchInteractive unfortunately only allows running command using the default shell. To allow for maintenance commands that run outside the systemd namespace, an escape hatch of some sort will need to be added to syschdemd.sh

sielicki:
Reviewing now.

Launcher/Launcher/src/Helpers/InstallationHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+        tarFail:
+        Console.Error.WriteLine("Could not find distro tarball");
+        return 1;
+    }
+
+    /// <summary>
+    ///     Unregister the distribution
+    /// </summary>
+    /// <returns>0 on success and an error code otherwise</returns>
+    public static int Uninstall() {
```
sielicki, line 83: Can we also add an option to nuke the directory?

sielicki (commented):
Looks great, awesome job.

Launcher/Launcher/src/Helpers/InstallationHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        if (WslApiLoader.WslIsDistributionRegistered(DistributionInfo.Name)) {
+            Console.Error.WriteLine($"{DistributionInfo.DisplayName} is already installed!");
+            return 0;
+        }
+        
+        const string tarFileName = "nixos-wsl-installer.tar.gz";
+
+        // Determine tarball location
+        var assemblyPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
```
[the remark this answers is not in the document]
nzbr: In the Release builds the tar file is included in the Msix package. That means, that it will always be at `../nixos-wsl-installer.tar.gz` relative to `Launcher.dll`. Using `Application.StartupPath` instead of the assembly location might be the better idea for debug builds

Launcher/Launcher-Appx/Package.appxmanifest
```diff
@@ -0,0 +1,59 @@
...
+        <Logo>Images\StoreLogo.png</Logo>
+    </Properties>
+
+    <Dependencies>
+        <TargetDeviceFamily Name="Windows.Universal" MinVersion="10.0.0.0" MaxVersionTested="10.0.0.0"/>
+        <TargetDeviceFamily Name="Windows.Desktop" MinVersion="10.0.14393.0" MaxVersionTested="10.0.14393.0"/>
+    </Dependencies>
+
+    <Capabilities>
+        <rescap:Capability Name="runFullTrust"/>
```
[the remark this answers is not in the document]
nzbr: The info in the appxmanifest will need to be adjusted with the publisher and package name for the Microsoft Store anyways
The [Microsoft docs](https://docs.microsoft.com/en-us/windows/uwp/publish) say that a privacy policy is "required for many apps". That means that we probably don't need one at all

nzbr:
Currently blocked by #138

puffnfresh:
Looks like this is unblocked now. Can we get a rebase for testing?

nzbr:
Feature-wise, this should be complete now! You can test it by downloading the `launcher` artifact from the GitHub Actions pipeline. You can use the `--distro-name` option with all commands to make sure it does not touch your existing NixOS install

Launcher/Launcher/src/Helpers/InstallationHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+        if (assemblyPath != null) {
+            tarPath = Path.Combine(assemblyPath, tarFileName);
+            if (File.Exists(tarPath)) return tarPath;
+        
+            // In the APPX package, the tarball is in the parent directory
+            var parentDirectory = Directory.GetParent(assemblyPath)?.FullName;
+            if (parentDirectory != null) {
+                tarPath = Path.Combine(parentDirectory, tarFileName);
+                if (File.Exists(tarPath)) return tarPath;
```
JamieMagee, line 95: nit

```suggestion
                if (File.Exists(tarPath))
                {
                    return tarPath;
                }
```

Launcher/Launcher/src/Helpers/InstallationHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+        var pwd = Directory.GetCurrentDirectory();
+        tarPath = Path.Combine(pwd, tarFileName);
+        if (File.Exists(tarPath)) return tarPath;
+        #endif
+        
+        // Look for the tarball in the same directory as the executable
+        var assemblyPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+        if (assemblyPath != null) {
+            tarPath = Path.Combine(assemblyPath, tarFileName);
+            if (File.Exists(tarPath)) return tarPath;
```
JamieMagee, line 89: nit

```suggestion
            if (File.Exists(tarPath))
            {
                return tarPath;
            }
```

Launcher/Launcher/Launcher.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net6.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+	    <RuntimeIdentifiers>win-x64;win-arm64</RuntimeIdentifiers>
+	    <AssemblyName>NixOS</AssemblyName>
+	    <Version>1.3.3.7</Version>
```
JamieMagee, line 10: Use a real version. Look at something like [`GitVersion`][1], [`Nerbank.Gitversioning`][2], or [`Minver`][3] if you don't want to manually manage it.

[1]: https://github.com/GitTools/GitVersion
[2]: https://github.com/dotnet/Nerdbank.GitVersioning
[3]: https://github.com/adamralph/minver

Launcher/Launcher/src/Helpers/StartupHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+                "sh -c \"exit\"", // Don't run anything, exit after we get a shell
+                false,
+                out var exitCode
+            );
+            if (exitCode != 0) {
+                goto fail;
+            }
+        } catch (WslApiException) {
+            goto fail;
+        }
```
JamieMagee, line 22: Can you log this exception message instead of a generic error message?

Launcher/Launcher/src/Helpers/InstallationHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /// <summary>
+    ///     Find the path of the installer tarball
+    /// </summary>
+    /// <returns>the full path to the tarball or null</returns>
+    public static string? FindTarball() {
+        const string tarFileName = "nixos-wsl-installer.tar.gz";
+
+        var tarPath = "";
```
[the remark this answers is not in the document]
nzbr: As far as I know, this shouldn't make a difference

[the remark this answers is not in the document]
JamieMagee: Yeah, I should have flagged this as a nit as well. I usually make this recommendation because it's more clear to any readers that you intended this to be an empty string.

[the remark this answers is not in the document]
nzbr: I've removed that line entirely, so this isn't applicable anymore

Launcher/Launcher/src/Helpers/StartupHelper.cs
nzbr: As you can see here [1], the Exception only contains the return code of the underlying WSL API function and no meaningful message. I have split the error handling now, so that it is clear, whether the command inside WSL or launching the command itself failed

[1]: https://github.com/nzbr/wsl.net/blob/main/src/WslApiException.cs

Launcher/Launcher/Launcher.csproj
nzbr: The `1.3.3.7` is just a placeholder, it gets replaced by the CI script. From what I understand, this may only be a numeric value, because otherwise I'd have used `DEV_BUILD` here

modules/installer.nix
```diff
@@ -35,6 +39,132 @@ with builtins; with lib; {
...
+                EXTERNAL=true
+              fi
+              cd $dir
+            fi
+
+            # Update channels
+            if ! $FLAKE && ! $EXTERNAL; then
+              echo "Updating channels..."
+              nix-channel --add https://nixos.org/channels/nixos-${config.system.nixos.release} nixos
+              nix-channel --update -v
```
SuperSandro2000, line 129: If we would add this to the installer it would not rely on the network, but also be bigger.

modules/installer.nix
```diff
@@ -35,6 +39,132 @@ with builtins; with lib; {
...
+              echo "Flake detected!"
+              echo "Please update your nixpkgs input to \"github:nixos/nixpkgs/nixos-${config.system.nixos.release}\" and run nixos-rebuild to complete the update"
+              echo "Press enter to continue..."
+              read
+              FLAKE=true
+            fi
+
+            # Check that nixos-rebuild builds the current configuration
+            EXTERNAL=false
+            if ! $FLAKE; then
```
SuperSandro2000, line 111: Why do we check this? Even if this fails I don't think it should abort the installation.

modules/installer.nix
```diff
@@ -35,6 +39,132 @@ with builtins; with lib; {
...
+            fi
+
+            # Unpack rootfs
+            echo "Unpacking..."
+            mkdir -p /tmp/nixos-wsl-updater/rootfs
+            ${pv}/bin/pv ${rootfs} | tar -C /tmp/nixos-wsl-updater/rootfs -xz
+
+            # Detect flake.nix
+            FLAKE=false
+            if [ -f /etc/nixos/flake.nix ]; then
```
SuperSandro2000, line 101: flakes can be anywhere and do not need to be here. Why do we need this?

Launcher/default.nix
```diff
@@ -0,0 +1,87 @@
...
+    dotnet-sdk
+    dotnetPackages.Nuget
+  ];
+
+  nugetDeps = linkFarmFromDrvs "${name}-nuget-deps" (map fetchNuGet [
+    { pname = "Microsoft.AspNetCore.App.Runtime.win-x64"; version = "6.0.0"; sha256 = "dacDLzjHNLXHIWdq+cBbXOfPpiqgHsWOp8cwrk+yDMk="; }
+    { pname = "Microsoft.NETCore.App.Host.win-x64"; version = "6.0.0"; sha256 = "3c1kDGpCrwlX858lyoE+u0fgmRgRwN692KqRN0ejJoA="; }
+    { pname = "Microsoft.NETCore.App.Runtime.win-x64"; version = "6.0.0"; sha256 = "RpVKg/7oiOSbPBNFgn1kt/+a55GRqv5QXtAO/+K0oY8="; }
+    { pname = "System.CommandLine"; version = "2.0.0-beta3.22114.1"; sha256 = "lcqMVesbt2E189e6KfG8HVmJc3hjo2ht48WnH1UaMkY="; }
+  ]);
```
SuperSandro2000, line 33: We should use the nixpkgs infrastructure for this

Launcher/default.nix
```diff
@@ -0,0 +1,87 @@
...
+
+  unpackPhase = ''
+    cp -r --no-preserve=mode $src/. .
```
SuperSandro2000, line 36: Are you searching for ``stripRoot = false;``?

Launcher/default.nix
```diff
@@ -0,0 +1,87 @@
...
+  '';
+
+  configurePhase = ''
+    runHook preConfigure
+
+    mkdir .home
+    export HOME=$PWD/.home
```
SuperSandro2000, line 43:
```suggestion
    export HOME=$TMPDIR
```
otherwise we could easily by mistake copy this to $out

Launcher/default.nix
```diff
@@ -0,0 +1,87 @@
...
+
+    export DOTNET_NOLOGO=1
+    export DOTNET_CLI_TELEMETRY_OPTOUT=1
```
SuperSandro2000, line 46: We don't need this with the nixpkgs functions for dotnet

Launcher/default.nix
```diff
@@ -0,0 +1,87 @@
...
+
+let
+  fetchNuGet = { pname, version, sha256 }: fetchurl {
+    name = "${pname}-${version}.nupkg";
+    url = "https://www.nuget.org/api/v2/package/${pname}/${version}";
+    inherit sha256;
+  };
+in
+stdenv.mkDerivation rec {
+  name = "NixOS-WSL-Launcher";
```
SuperSandro2000, line 19: we should set pname+version

Launcher/default.nix
```diff
@@ -0,0 +1,87 @@
+{ pkgs ? import <nixpkgs> { }
+, stdenv ? pkgs.stdenv
+, lib ? pkgs.lib
+, writeText ? pkgs.writeText
+, dotnet-sdk ? pkgs.dotnet-sdk
+, dotnetPackages ? pkgs.dotnetPackages
+, fetchurl ? pkgs.fetchurl
+, linkFarmFromDrvs ? pkgs.linkFarmFromDrvs
+}:
```
SuperSandro2000, line 9: Can't we rely on callPackage?

Launcher/Launcher/src/Commands/Update.cs
```diff
@@ -0,0 +1,60 @@
...
+                    DistributionInfo.Name,
+                    $"sudo {tmpDir}/nix/wsl-installer/updater.sh",
+                    false,
+                    out exitCode
+                );
+                
+                if (exitCode != 0) {
+                    Console.Error.WriteLine("\nAn error occured in the update script");
+                } else {
+                    Console.WriteLine("\nUpdate finished successfully");
```
SuperSandro2000, line 49: Does this log the original error?

Launcher/Launcher/src/Commands/Update.cs
```diff
@@ -0,0 +1,60 @@
...
+                if (!StartupHelper.BootDistro()) return 1;
+
+                var tarball = InstallationHelper.FindTarball();
+                if (tarball == null) return 1;
+
+                var path = Path.GetDirectoryName(tarball);
+                if (path == null) return 1;
+                Directory.SetCurrentDirectory(path);
+
+                var tmpDir = "/tmp/nixos-wsl-updater/installer";
```
SuperSandro2000, line 24: We should rm that, just in case it exists or use mktmp

.github/workflows/release.yml
```diff
@@ -19,9 +19,17 @@ jobs:
         with:
           name: installer
 
+      - name: Download appx 📥
+        uses: actions/download-artifact@v3
+        with:
+          name: appx
+
+      - name: Rename appx ✏️
+        run: mv */*.msixbundle NixOS-WSL.msixbundle
```
SuperSandro2000, line 28: actions do not allow post run or similar?

.github/workflows/main.yml
```diff
@@ -47,24 +60,86 @@ jobs:
...
+        with:
+          name: launcher
+          path: Launcher/Launcher/pkg
+
+      - name: Build AppX package
+        run: cd Launcher; msbuild Launcher-Appx\Launcher-Appx.wapproj /p:Configuration=Release /p:Platform=x64 /p:ContinuousIntegrationBuild=true /p:Deterministic=true
+
+      - name: Sign package
+        env:
+          APPX_SIGNING_CERT: ${{ secrets.APPX_SIGNING_CERT }}
```
SuperSandro2000, line 122: Can we store that in sops or similar so that we can retrieve it again? secrets are write only in the settings.

.github/workflows/main.yml
```diff
@@ -47,24 +60,86 @@ jobs:
...
+          vs-version: '[17.2,'
+
+      - name: Set Versions
+        env:
+          NIXOS_WSL_VERSION: ${{ needs.prepare.outputs.version }}
+        run: |
+          ((Get-Content -path .\Launcher\Launcher\Launcher.csproj -Raw) -replace '1.3.3.7',$env:NIXOS_WSL_VERSION) | Set-Content -Path .\Launcher\Launcher\Launcher.csproj
+          ((Get-Content -path .\Launcher\Launcher-Appx\Package.appxmanifest -Raw) -replace '1.3.3.7',$env:NIXOS_WSL_VERSION) | Set-Content -Path .\Launcher\Launcher-Appx\Package.appxmanifest
+      - name: Build Launcher
+        run: cd Launcher; dotnet publish --self-contained -r win-x64 -c Release -p:ContinuousIntegrationBuild=true -p:Deterministic=true --output Launcher\pkg Launcher
```
SuperSandro2000, line 105:
```suggestion
        run: |
          cd Launcher
          dotnet publish --self-contained -r win-x64 -c Release -p:ContinuousIntegrationBuild=true -p:Deterministic=true --output Launcher\pkg Launcher
```

.github/workflows/main.yml
```diff
@@ -47,24 +60,86 @@ jobs:
...
+
+      - name: Download installer
+        uses: actions/download-artifact@v3
+        with:
+          name: installer
+
+      - name: Setup MSBuild
+        uses: microsoft/setup-msbuild@v1.1
+        with:
+          vs-version: '[17.2,'
```
SuperSandro2000, line 96: Is this a version regex? Looks half complete to me.

.github/workflows/main.yml
```diff
@@ -33,8 +34,20 @@ jobs:
         run: |
           nix-instantiate --json --eval --strict -E 'with builtins; attrNames (getFlake (toString ./.)).checks.${currentSystem}' | perl -pe 's|(.*)|checks=\1|' >>$GITHUB_OUTPUT
 
+      - name: Generate Version 🏷️
+        id: version
+        run: |
+          TAG_COUNT=$(git rev-list --tags --no-walk --count)                                                                 # Count all tags
```
SuperSandro2000, line 40: Not sure if we need that at all. If we ever delete a tag this will be off.

.github/workflows/main.yml
```diff
@@ -33,8 +34,20 @@ jobs:
...
+          COMMIT_COUNT=$(git rev-list --use-bitmap-index --count $(git rev-list --tags --no-walk --max-count=1)..HEAD)       # Count all commits since the last tag
```
SuperSandro2000, line 41: Same with this, also since we probably bump flake.nix a lot this number is rather infalted.

SuperSandro2000 (commented):
didn't look at all the c# files.

Can we add one final new line to all files so that git is happy or will Windoof dislike that?

.github/workflows/main.yml
nzbr: It's a range from (including) 17.2 to infinity

.github/workflows/main.yml
nzbr: I can't even edit the repo settings (or commit to it through git for that matter), so having this updateable _somewhere_ else is probaly a good idea, yes

.github/workflows/release.yml
nzbr: What do you mean by that?

Launcher/Launcher/src/Commands/Update.cs
nzbr: I'm not really happy with the update process in general, but I haven't found a nicer way to handle it than unpacking the installer somewhere and subsequently unpacking the rootfs

mktemp sounds like a good idea here

Launcher/default.nix
nzbr: This doesn't get called from the flake. If it was, i'd say yes. I haven't managed to build the APPX bundle inside nix (or on linux in general), so this whole file isn't even used in the CI at the moment. We could probably remove this altogether

nzbr: Feel free to give it a shot though, if you want to try to build it

modules/installer.nix
nzbr: Having a flake here specifically means, that `nixos-rebuild` will use it even without specifying it through the `--flake` option. So the below check wouldn't catch it

modules/installer.nix
nzbr: It checks that `nixos-rebuild` will build the config you are currently running, and if it doesn't, it skips updating the nixpkgs channel and the final `nixos-rebuild` steps and tells you to perform it manually. The install does not get cancelled

modules/installer.nix
nzbr: Given that this is for updating, it should always give you the latest nixpkgs. Trying to update without access to the internet would break at nixos-rebuild anyway

Launcher/Launcher/src/Commands/Update.cs
nzbr: WslLaunchInteractive attaches the standard io, so everything the updater.sh script logs will be visible

.github/workflows/main.yml
nzbr: True, but I don't see a better way here. The relevant code is already on main (#183). Not being able to delete tags is something I'd be fine with, honestly. If you have a better Idea, we can still change it though

nzbr:
That was a lot 😅 Thank you for the effort!

nzbr:
So apparently the formatter made the appxmanifest invalid. _that's great_

Launcher/Launcher/src/Commands/Update.cs
nzbr: I have now chosen a combined approach: the directory name gets generated from the version number and `rm`ed before the unpack

modules/installer.nix
SuperSandro2000: Can't we just instruct people to manually do the update?
Most releases we have deprecations that people should handle anyway and I suspect this to open another pandoras box.

modules/installer.nix
SuperSandro2000: but I could place my flake also under /some/random/directory and point nixos-rebuild to that.

.github/workflows/release.yml
SuperSandro2000: That we do the mv command after the previous and not in a new step. Not that important.

modules/installer.nix
nzbr: That's what the other check (the one that runs `nixos-rebuild build`) is for. If `nixos-rebuild` without the `--flake` attribute does not produce the current config, it will not run it and prompt you to update your flake inputs instead. If there's a flake in `/etc/nixos/flake.nix`, it will print a slightly different prompt, but in both cases, neither will the channels be changed, not `nixos-rebuild switch` run

.github/workflows/release.yml
nzbr: I don't think something like that exists, and it doesn't really make a difference either way (unlike with dockerfiles for example)

modules/installer.nix
nzbr: Having people try to run this on NixOS versions that are older than what we test against might lead to new and interesting bugs that I don't want to deal with. There should probably be a check for nixos-unstable though

modules/installer.nix
nzbr: Because the two checks have the same effect anyway (except for the message they print), it might make sense to combine them to make the script shorter

.github/workflows/main.yml
nzbr: Everyone using this is running at least one Microsoft product and should therefore be accustomed to ludicrously high version numbers like `10.0.22623.1245`

Jokes aside: I don't think that the commit count will ever exceed three (or at the very most four) digits and that's totally acceptible

JamieMagee (approved):

JamieMagee:
@nzbr I'd love to see this in the store. Let me know how I can help.

nzbr:
There's gonna be changes to the Installation and Update procedures, which will require adjustments to the launcher's code. If I remember correctly, the basic features were all in place, so I don't think there's anything you could help with right now. Thanks for the offer though

nzbr:
I moved the branch to this repo and will therefore need to re-open the PR

PR #96, opened 2022-06-01
Enable opengl support

nzbr:
This adds support for running applications that use OpenGL with WSLg.
See also: #88


grantbevis (approved):

grantbevis:
@nzbr Looks good mate

PR #102, opened 2022-06-08
Mention Windows 10 explicitly

SuperSandro2000:


modules/wsl-distro.nix
```diff
@@ -52,7 +52,7 @@ with builtins; with lib;
 
             Default (null): autodetect, at some performance cost.
             To avoid the performance cost, set this to true for WSL Preview 0.58 and up,
-            or to false for older versions (including pre-Microsoft Store).
+            or to false for any older versions, including pre-Microsoft Store used on Windows 10.
```
nzbr, line 55: I think this should be "and Windows 10" instead. That makes it more clear that pre-MS Store is not a Windows 10 only thing. W11 also ships with an older version, that can then be updated through the Store (although that may change in the future)

SuperSandro2000:
```suggestion
            or to false for any older versions, including pre-Microsoft Store and Windows 10.
```

PR #105, opened 2022-06-10
Don't add the default user to docker group by default

### LINKED ISSUE #101: docker-native should not always add the default user to the docker group
Adding a user to the docker group basically gives the user root rights withou a password. The option at https://github.com/nix-community/NixOS-WSL/blob/main/modules/docker-native.nix#L26-L28 should at least be made configurable.

SuperSandro2000:
Closes #101

### DIFF
--- a/modules/docker-native.nix
+++ b/modules/docker-native.nix
@@ -3,6 +3,16 @@ with builtins; with lib; {
 
   options.wsl.docker-native = with types; {
     enable = mkEnableOption "Native Docker integration in NixOS.";
+
+    addToDockerGroup = mkOption {
+      type = bool;
+      default = false;
+      description = ''
+        Wether to add the default user to the docker group.
+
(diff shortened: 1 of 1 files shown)


nzbr:
The default for `addToDockerGroup` should probably be `config.security.sudo.wheelNeedsPassword`, because by default the user does not have a password anyway

modules/docker-native.nix
```diff
@@ -3,6 +3,16 @@ with builtins; with lib; {
 
   options.wsl.docker-native = with types; {
     enable = mkEnableOption "Native Docker integration in NixOS.";
+
+    addToDockerGroup = mkOption {
+      type = bool;
+      default = false;
```
SuperSandro2000, line 9:
```suggestion
      default = config.security.sudo.wheelNeedsPassword;
```

PR #117, opened 2022-07-11
docker-desktop: fix bind mounts

nzbr:
The `/mnt` mount gets marked as shared in the systemd wrapper script, so that the bind mounts created by the docker desktop proxy are present outside the systemd namespace

See also #89 #90

@hschaeidt

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -61,7 +61,7 @@ with builtins; with lib;
   config =
     let
       cfg = config.wsl;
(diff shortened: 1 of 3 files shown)


nzbr:
Can someone verify, that this patch works for people other than me
For example `docker run --rm -it -v $PWD:/pwd alpine ls -la /pwd` should show the contents of the current directory, while listing an empty directory without the patch

pogopaule:
Does not work for me.

```
❯ docker run --rm -it -v $PWD:/pwd alpine ls -la /pwd
total 4
drwxr-xr-x    2 root     root            40 Oct 24 12:01 .
drwxr-xr-x    1 root     root          4096 Oct 24 12:01 ..
```

See https://github.com/nix-community/NixOS-WSL/discussions/148

PR #119, opened 2022-07-15
interop: move to its own module and disable re-registration by default

nzbr:
Due to the amount of side-effects that touching the binfmt registrations introduces (no interop in other distros, broken WSLg, etc)  it is now disabled by default. When the configuration contains another binfmt registration (and would therefore break interop) a warning is issued during evaluation

@K900

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,10 +17,11 @@
       nixosModules.wsl = {
         imports = [
           ./modules/build-tarball.nix
-          ./modules/wsl-distro.nix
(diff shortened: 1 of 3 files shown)


K900 (approved):
I don't really like this conceptually, but I don't think we can really fix it without help from upstream.

The actual diff LGTM.

nzbr:
Once https://github.com/microsoft/WSL/issues/8203 is resolved, we can enable it by default again. This is just meant as a temporary fix

PR #125, opened 2022-08-21
Store corruption fixes

K900:
systemd tries to create the autovt@ unit on startup. Make it not do that, and also make it so we can't fuck it up again.


scripts/syschdemd.nix
```diff
@@ -0,0 +1,54 @@
...
+    coreutils
+    daemonize
+    glibc  # getent
+    gnugrep
+    systemd  # machinectl
+    util-linux  # nsenter
+    wrapper
+  ];
+  inherit defaultUser defaultUserHome;
+}
```
SuperSandro2000, line 54: Missing a final new line which is important that git can generate diffs properly.

scripts/syschdemd.nix
```diff
@@ -0,0 +1,54 @@
...
+  coreutils, 
+  daemonize,
+  glibc,
+  gnugrep,
+  systemd, 
+  util-linux, 
+  
+  defaultUser,
+  defaultUserHome
+}:
```
SuperSandro2000, line 15: Can we use nixpkgs-fmt instead of alejandra?

scripts/syschdemd.nix
```diff
@@ -0,0 +1,54 @@
...
+let
+  writeShellScript = { name, src, path, ... }@args:
+    stdenvNoCC.mkDerivation ({
```
SuperSandro2000, line 18: This could be replaced with runCommand

scripts/wrapper.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+mount -t binfmt_misc binfmt_misc /proc/sys/fs/binfmt_misc
+
+exec systemd
```
SuperSandro2000, line 6: also missing a final new line

scripts/syschdemd.nix
K900: This is actually formatted manually

PR #126, opened 2022-08-22
feat: rewrite syschdemd

K900:
Mostly just a big pile of cleanups.

Important changes:
- refactor and clean up the code, fix all the shellcheck complaints
- stop trying to guess paths in scripts and wrap them correctly instead
- use nsfs instead of trying to figure out the right PID to copy namespaces from
- clean up $WSLPATH to remove extra impure entries
- don't try to restart systemd if it died
- make sure the store is read-only before we do anything, so systemd can't mess with it


scripts/syschdemd.nix
```diff
@@ -0,0 +1,51 @@
...
+, lib
+, coreutils
+, daemonize
+, glibc
+, gnugrep
+, systemd
+, util-linux
+, defaultUser
+, automountPath
+,
```
SuperSandro2000, line 12:
```suggestion
```

scripts/syschdemd.nix
```diff
@@ -0,0 +1,51 @@
...
+}:
+let
+  writeShellScript =
```
SuperSandro2000, line 15: This shadows a function in nixpkgs https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/trivial-builders.nix#L213

scripts/syschdemd.sh
```diff
@@ -0,0 +1,134 @@
...
+        machinectl \
+            --quiet \
+            --uid=@defaultUser@ \
+            --setenv=INSIDE_NAMESPACE=true \
+            --setenv=WSLPATH="$(clean_wslpath)" \
+            shell .host \
+                /bin/sh -c "cd \"$PWD\"; $exportCmd; source /etc/set-environment; exec $command"
+}
+
+main "$@"
```
SuperSandro2000, line 134: Missing final new line

scripts/syschdemd.sh
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+[ "${NIXOS_WSL_DEBUG:-}" == "1" ] && set -x
```
SuperSandro2000, line 4:
```suggestion
[[ -v NIXOS_WSL_DEBUG ]] && set -x
```
testing if it is set to anything maybe makes it easier if people accidentally use true

scripts/syschdemd.nix
K900: Fixed.

scripts/syschdemd.sh
K900: I just reformatted the whole thing with shfmt.

SuperSandro2000: Just for my interest: What settings are you using? Maybe we should document that somewhere to not reformat files multiple times. I am always using ``shfmt -ci -i 2 -s -w $FILE``.

K900: I did add it as a check in CI.

nzbr:
Docker Desktop is broken, because the `mount --make-rshared ${automountPath}` line is missing from the systemd wrapper script. It's needed so that the bind mount that docker creates is present outside of the systemd container

nzbr:
Other than that, everything seems to be working

K900:
I tried adding it to the _outside_ script, can you see if that works?

K900:
I am very confused by how the mounts work then. Can you run `findmnt -o+PROPAGATION` outside the container?

K900:
Is this from inside the namespace or outside?

nzbr:
The command output is from inside the systemd namespace, but not the docker container. Outside the systemd namespace it looks like this:

[mounts-outside.txt](https://github.com/nix-community/NixOS-WSL/files/9426267/mounts-outside.txt)

nzbr:
This is also a problem for other mounts created inside the namespace: I have a tmpfs on /tmp in my NixOS installation (with the current syschdemd). Because of that any files in /tmp can't be accessed from outside it (e.g. through SMB with the explorer or interop) and trying to use interop in there results in a not at all helpful error:
image

When I made the docker-desktop fix, I also tried removing mount namespacing completely, but that means that the namespace and the ouside have the same `/proc` filesystem which breaks even more things

K900:
Can you try the latest commit? I think unshare was just doing a weird.

nzbr:
The latest commit seems to be broken
image

K900:
Forgot to remove the other thing. Can you try it now?

nzbr:
image

Seems to be working now

K900:
Squashed, should be good to go now.

PR #134, opened 2022-09-23
feat: native systemd support

K900:
Known issues:
- user services don't really work (https://github.com/microsoft/WSL/issues/8918, https://github.com/microsoft/WSL/issues/8842)


K900:
Workaround for user sessions: `sudo loginctl enable-linger $USER`

K900:
See also: https://github.com/microsoft/WSL/issues/8842

nzbr:
For logging, you could use something like
```bash
.../activate |& logger --socket-errors=off -st activate 2>/dev/kmsg
```
to make the log output appear in dmesg
Systemd can be started with `--log-target=kmsg` to achieve that as well (though I haven't tested that)

Issue #87, opened 2022-04-27
nsenter: failed to parse pid: '-p'


Panke:
## Bug description
I installed it but I cannot start it, because `nsenter: failed to parse pid: '-p'.

## To Reproduce
1. Get the installer and follow readme instructions (see terminal output below)

Include relevant console logs

```
PS C:\Users\bcr88\Downloads> wsl --import NixOS ..\NixOS\ .\nixos-wsl-installer.tar.gz --version 2
PS C:\Users\bcr88\Downloads> wsl -d NixOS
Unpacking root file system...
 184MiB 0:00:07 [24.6MiB/s] [=================================================================================================================================================================>] 100%
Activating nix configuration...
Copying /usr/share/applications
Copying /usr/share/icons
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
setting up /etc...
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
Cleaning up installer files...
finding garbage collector roots...
removing stale link from '/nix/var/nix/gcroots/auto/6yfbrv8wfkyl40wq6827lamr9m29z7wg' to '/build/nix/var/nix/profiles/per-user/root/channels-1-link'
removing stale link from '/nix/var/nix/gcroots/auto/zxi8v969la5n36bg7gs016wgcif4qjhp' to '/build/nix/var/nix/profiles/system-1-link'
deleting garbage...
deleting '/nix/store/svw89j47vr3h792f75nfgh5x9s48rn0w-pv-1.6.20'
deleting '/nix/store/m0qmllywgy9wl4ia7vvpz7p53kzxv681-busybox-1.34.1'
deleting '/nix/store/m7cghw288wks5s5mkxpg1k3h5ivlg1ch-installer.sh'
deleting '/nix/store/8hm7r6m8y0fjrb4qjv8gs1ba04g51166-tarball'
deleting '/nix/store/y608sw04xdqhha69raxpj2rsk007n7wf-wsl-prepare'
deleting '/nix/store/4ryxif8fmjb362ry8ylx29hsxp0zgpax-source'
[...]

PR #143, opened 2022-10-13
Fix install instructions

SuperSandro2000:
If the other archive is used, you end up with #87

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ First, [download the latest release\'s installer](https://github.com/nix-communi
 Then open up a Terminal, PowerShell or Command Prompt and run:
 
 ```sh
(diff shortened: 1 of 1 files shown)


nzbr:
Directly importing the rootfs leads to an "Unspecified Error" if it contains _any_ hard link, because Microsoft's tar implementation apparently just crashes when it encounters a hard link. That's why the installer exists in the first place. The installer tarball is packed with an option that replaces all hard links with symlinks

The re-written `syschdemd.sh` does not seem to have the problem from #87, so maybe creating a new release would be a better idea

SuperSandro2000:
I was installing nixos-wsl on another windows machine and using the installer loaded but then just hang midway and a new terminal couldn't be open. Using the rootfs worked.

Issue #138, opened 2022-10-04
wsl.exe always returns 0 exit code


nzbr:
## Bug description

Regardless of the exit code of the command that runs inside WSL, the `wsl.exe` binary always returns an exit code of `0`

## To Reproduce
Run
```pwsh
wsl.exe -d NixOS false
echo $LASTEXITCODE
```

**Expected behavior**

`$LASTEXITCODE` should be `1`

## Logs
```
PS C:\Users\nzbr> wsl -d NixOS false

PS C:\Users\nzbr> echo $LASTEXITCODE
0
PS C:\Users\nzbr> wsl -d Ubuntu false
PS C:\Users\nzbr> echo $LASTEXITCODE
1
PS C:\Users\nzbr>

```

I verified that it works with Ubuntu, so this is most likely a problem with syschdemd. I haven't tested with the new native systemd feature yet, but I expect that to work

nzbr:
Apparently, we need to switch back to systemd-run instead of machinectl in order to get this to work
https://github.com/systemd/systemd/issues/4215

PR #144, opened 2022-10-13
Do not import minimal profile by default

SuperSandro2000:
- `environment.noXlibs` breaks often things and is a hassle to use
- `documentation.enable` having documentation available is desired
- `documentation.nixos.enable` removes the man page for nixos-rebuild which is triggered by --help
- `programs.command-not-found.enable` also command-not-found is useful

see https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/profiles/minimal.nix

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -6,8 +6,6 @@ let
 in
 {
   imports = [
-    "${modulesPath}/profiles/minimal.nix"
-
     nixos-wsl.nixosModules.wsl
   ];
 


SuperSandro2000:
github action is being flaky

nzbr:
Good idea! We should remove `boot.isContainer` as well due to similar quirks. Doing that also seems to help in solving #138, because isContainer changes a PAM module that's required for that to work from `required` to `optional`

A working replacement looks like this (in wsl-distro.nix):
```nix
      # We don't need a boot loader
      boot.loader.grub.enable = false;
      system.build.installBootLoader = pkgs.writeShellScript "fake-bootloader" "";
      boot.initrd.enable = false;
      system.build.initialRamdisk = pkgs.runCommand "fake-initrd" { } ''
        mkdir $out
        touch $out/${config.system.boot.loader.initrdFile}
      '';
      system.build.initialRamdiskSecretAppender = pkgs.writeShellScriptBin "append-initrd-secrets" "";
```

SuperSandro2000:
Lets move the removal of `boot.isContainer` to another PR to not open that pandoras box right now. For example I can imagine that we get problems with [udev which gets activated then](https://github.com/NixOS/nixpkgs/blob/0f3ea348755ec3370f5854568ed890789e4fb537/nixos/modules/services/hardware/udev.nix#L348), audit, [modprobe](https://github.com/NixOS/nixpkgs/blob/0f3ea348755ec3370f5854568ed890789e4fb537/nixos/modules/system/boot/modprobe.nix#L41)

PR #145, opened 2022-10-17
remove boot.isContainer

nzbr:
Do not set boot.isContainer to true, but set the required options directly. This gets around some quirks isContainer introduces in other modules (especially one in PAM that is related to solving #138). This might introduce issues related to `udev`, `audit` and `modprobe`, which get activated by removing that option; though i have not encountered any so far

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -50,10 +50,16 @@ with builtins; with lib;
         };
       };
 
-      # WSL is closer to a container than anything else
-      boot.isContainer = true;
+      # We don't need a boot loader
+      boot.loader.grub.enable = false;
+      system.build.installBootLoader = pkgs.writeShellScript "fake-bootloader" "";
+      boot.initrd.enable = false;
+      system.build.initialRamdisk = pkgs.runCommand "fake-initrd" { } ''
+        mkdir $out
+        touch $out/${config.system.boot.loader.initrdFile}
+      '';
+      system.build.initialRamdiskSecretAppender = pkgs.writeShellScriptBin "append-initrd-secrets" "";
 
-      environment.noXlibs = lib.mkForce false; # override xlibs not being installed (due to isContainer) to enable the use of GUI apps
       hardware.opengl.enable = true; # Enable GPU acceleration
 
       environment = {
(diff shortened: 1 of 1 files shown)


modules/wsl-distro.nix
```diff
@@ -50,10 +50,16 @@ with builtins; with lib;
         };
       };
 
-      # WSL is closer to a container than anything else
-      boot.isContainer = true;
+      # We don't need a boot loader
+      boot.loader.grub.enable = false;
+      system.build.installBootLoader = pkgs.writeShellScript "fake-bootloader" "";
```
SuperSandro2000, line 55:
```suggestion
      system.build.installBootLoader = "${pkgs.coreutils}/bin/true"
```
isContainer solved it like this which is a bit shorter and one derivation less

modules/wsl-distro.nix
```diff
@@ -50,10 +50,16 @@ with builtins; with lib;
...
+      boot.initrd.enable = false;
+      system.build.initialRamdisk = pkgs.runCommand "fake-initrd" { } ''
+        mkdir $out
+        touch $out/${config.system.boot.loader.initrdFile}
+      '';
+      system.build.initialRamdiskSecretAppender = pkgs.writeShellScriptBin "append-initrd-secrets" "";
```
SuperSandro2000, line 61: Maybe we can use true here, too?

nzbr: We can't, because the binary needs to be named `append-initrd-secrets`. We could in theory put a symlink to `true` there, but I don't think that would make any difference while being a longer nix expression

nzbr:
I haven't encountered any problems so far, so I'm just merging it

SuperSandro2000:
Now we also [activated console](https://github.com/NixOS/nixpkgs/blob/9331485549ae5ab046a7ee681ccd4463deb83226/nixos/modules/config/console.nix#L38), linux firmware and kernel which is not working in WSL.

PRs to make this better:
- https://github.com/NixOS/nixpkgs/pull/197867


so we now should set:

```
hardware.console = false;
```

and probably take a look if the following units can be removed:

```+ lvm2-2.03.16
+ unit-console-getty.service-disabled
+ unit-post-resume.service
+ unit-post-resume.target
+ unit-pre-sleep.service
+ unit-systemd-modules-load.service
+ unit-systemd-timesyncd.servic
```

SuperSandro2000:
and to add:
- https://github.com/NixOS/nixpkgs/pull/197917

PR #147, opened 2022-10-19
NixOS users.users.<user> attribute name and .name can differ

KoviRobi:
This is admittedly a niche use case, they are the same most of the time, but on my computer for historical purposes I have `users.users.default-user` which has a different `.name` depending on whether Nix is building my work configuration or uni configuration (hence why it's `default-user`).


nzbr:
Looks good. Thank you!

SuperSandro2000:
Before always including the wsl module worked but now it errors:
```
building the system configuration...
error: attribute 'nixos' missing

at /nix/store/pqfb6jic287q3d7s9jvz1w8lgrjnm3bi-source/modules/wsl-distro.nix:40:23:

39|         inherit (cfg) automountPath;
40|         defaultUser = config.users.users.${cfg.defaultUser};
|                       ^
41|       };
```

KoviRobi:
I guess you don't have a `config.users.users.nixos`? But https://github.com/nix-community/NixOS-WSL/blob/27679db0469bc51ce4d0b334c74c4b0642d62009/modules/wsl-distro.nix#L98 should get you that. I'll have to dig a bit deeper

SuperSandro2000:
Yes, I set ``wsl.defaultUser = "sandro"``. I think you cannot use the user done here in the same config section.

KoviRobi:
Hmm I've tried creating a simple test case at https://github.com/KoviRobi/nixos-wsl-test and the following works
```
nix build github:KoviRobi/nixos-wsl-test\#nixosConfigurations.wsl-test.config.system.build.toplevel
```
Also
```console
nix-repl> :lf github:KoviRobi/nixos-wsl-test
Added 10 variables.

nix-repl> builtins.attrNames nixosConfigurations.wsl-test.config.users.users
[ "messagebus" "nixbld1" "nixbld10" "nixbld11" "nixbld12" "nixbld13" "nixbld14" "nixbld15" "nixbld16" "nixbld17" "nixbld18" "nixbld19" "nixbld2" "nixbld20" "nixbld21" "nixbld22" "nixbld23" "nixbld24" "nixbld25" "nixbld26" "nixbld27" "nixbld28" "nixbld29" "nixbld3" "nixbld30" "nixbld31" "nixbld32" "nixbld4" "nixbld5" "nixbld6" "nixbld7" "nixbld8" "nixbld9" "nobody" "nscd" "root" "systemd-coredump" "systemd-journal-gateway" "systemd-network" "systemd-oom" "systemd-resolve" "systemd-timesync" "test" ]
```
has no user `nixos` only `test`

SuperSandro2000:
Found the problem. Setting ``wsl.enable = false`` is still evaluating some WSL parts which want to access the created user https://github.com/nix-community/NixOS-WSL/blob/main/modules/wsl-distro.nix#L150-L184

SuperSandro2000:
#151

PR #153, opened 2022-11-02
wsl.conf: proper option types

nzbr:
Added proper types for all options that can be set in wsl.conf to avoid typos and using wrong types

### DIFF
--- a/modules/wsl-conf.nix
+++ b/modules/wsl-conf.nix
@@ -1,0 +1,90 @@
+{ lib, pkgs, config, ... }:
+with builtins; with lib; {
+
+  imports = [
+    (mkRenamedOptionModule [ "wsl" "automountPath" ] [ "wsl" "wslConf" "automount" "root" ])
+    (mkRenamedOptionModule [ "wsl" "automountOptions" ] [ "wsl" "wslConf" "automount" "options" ])
+  ];
+
+  # See https://learn.microsoft.com/en-us/windows/wsl/wsl-config#configuration-settings-for-wslconf for all options
+  options.wsl.wslConf = with types; {
+    automount = {
+      enabled = mkOption {
+        type = bool;
+        default = true;
+        description = "Automatically mount windows drives under ${config.wsl.wslConf.automount.root}";
+      };
+      mountFsTab = mkOption {
+        type = bool;
+        default = false;
+        description = "Mount entries from /etc/fstab through WSL. You should probably leave this on false, because systemd will mount those for you.";
+      };
+      root = mkOption {
+        type = str;
+        default = "/mnt/";
+        description = "The directory under which to mount windows drives.";
+      };
+      options = mkOption {
+        type = commas; # comma-separated strings
+        default = "metadata,uid=1000,gid=100";
+        description = "Comma-separated list of mount options that should be used for mounting windows drives.";
+      };
+    };
+    boot = {
+      command = mkOption {
+        type = str;
+        default = "";
+        description = "A command to run when the distro is started.";
+      };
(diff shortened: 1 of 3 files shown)


modules/wsl-conf.nix
```diff
@@ -0,0 +1,90 @@
+{ lib, pkgs, config, ... }:
+with builtins; with lib; {
+
```
SuperSandro2000, line 3:
```suggestion

with lib; {
```
almost all builtins are inherited into lib.

modules/wsl-conf.nix
```diff
@@ -0,0 +1,90 @@
...
+      };
+    };
+    interop = {
+      enabled = mkOption {
+        type = bool;
+        default = true;
+        description = "Support running Windows binaries from the linux shell.";
+      };
+      appendWindowsPath = mkOption {
+        type = bool;
```
SuperSandro2000, line 52: This is currently a string which contains the boolean.

```
error: A definition for option `wsl.wslConf.interop.appendWindowsPath' is not of type `boolean'. Definition values:
       - In `/nix/store/byrkin8wmr9hbzqhnsfa01bj2g7xij81-source/common/modules.nix': "false"
```

modules/wsl-conf.nix
```diff
@@ -0,0 +1,90 @@
...
+    boot = {
+      command = mkOption {
+        type = str;
+        default = "";
+        description = "A command to run when the distro is started.";
+      };
+      systemd = mkOption {
+        type = bool;
+        default = false;
+        description = "Use systemd as init. There's no need to enable this manually, use the wsl.nativeSystemd option instead";
```
SuperSandro2000, line 42: We should assert or warn this.

modules/wsl-conf.nix
```diff
@@ -0,0 +1,90 @@
...
+      options = mkOption {
+        type = commas; # comma-separated strings
+        default = "metadata,uid=1000,gid=100";
+        description = "Comma-separated list of mount options that should be used for mounting windows drives.";
+      };
+    };
+    boot = {
+      command = mkOption {
+        type = str;
+        default = "";
```
SuperSandro2000, line 36: example would be great.

modules/wsl-conf.nix
nzbr: Didn't know that works - thanks

modules/wsl-conf.nix
nzbr: This isn't really useful on NixOS, because we have systemd. The command is started using the root shell, so syschdemd would drop it to user privileges anyway. MS intends this to be used to start a daemon in the background (their example is starting docker)

modules/wsl-conf.nix
nzbr: Where is that string? Was it an old default config? The current default configuration.nix as well as my flake config build just fine

modules/wsl-conf.nix
nzbr: Good idea, I'll add a warning

modules/wsl-conf.nix
SuperSandro2000: It is some option I have currently set but I no longer know why and from where I got it. Also can't find any reference in code.

SuperSandro2000: I think I now know the root cause of this: before wslConf was a freeform setting that accepted a lot more than it should. Now that the options are properly typed errors are properly surfaced and you can no longer set an identical string to an otherwise boolean option.

SuperSandro2000 (approved):
didn't test again but looks alright

PR #166, opened 2022-11-21
Fix permissions for files in /dev/pts

SuperSandro2000:
The pts should have $USER:tty permissions, not root:tty. This broke gpg with the following not very helpful error message:
gpg: signing failed: Permission denied


nzbr:
I have seen your PR, but I wanted to finish the ci tests first, so I can run it against them once they are merged

PR #172, opened 2022-11-30
Fix X11 socket symlink creation

SuperSandro2000:
by updating the tmpfile.d rule to match the one created by upstream WSL

Or do we just want to remove this?


nzbr:
If this _works_, I don't see a reason not to merge it

SuperSandro2000:
but I don't want to continuously update this. Maybe we should really just drop it and let MS handle this until it breaks again.

PR #175, opened 2022-12-09
Use upstream settings instead of workarounds to disable unsupported settings

### LINKED ISSUE #149: Cleanup after removing boot.isContainer
Follow up to https://github.com/nix-community/NixOS-WSL/pull/145

ToDos right now:
- [x] disable timesyncd
- [x] add ``powerManagement.enable = false;`` to disable the following units:
```
+ unit-post-resume.service
+ unit-post-resume.target
+ unit-pre-sleep.service
```

Wait for the following PRs:
- https://github.com/NixOS/nixpkgs/pull/197867
- https://github.com/NixOS/nixpkgs/pull/197870
- https://github.com/NixOS/nixpkgs/pull/197917
- https://github.com/NixOS/nixpkgs/pull/197922
- https://github.com/NixOS/nixpkgs/pull/197925

ToDos after that:
- set the following:
```
boot.modprobeConfig.enable = false; # maybe not required after patching top-level activation?
boot.kernel.enable = false;
console.enable = false;
hardware.console = false;
services.udev.enable = false;
```
- revert/fix workarounds introduced in #145
- cleanup nixos/modules/virtualisation/container-config.nix

other questions:

sshd.pam changed. Is that important?

```diff
-session optional pam_loginuid.so
+session required pam_loginuid.so
```

SuperSandro2000:
Closes #149

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -37,20 +37,20 @@ with lib; {
     mkIf cfg.enable (
       mkMerge [
         {
-          # We don't need a boot loader
-          boot.loader.grub.enable = false;
-          system.build.installBootLoader = "${pkgs.coreutils}/bin/true";
-          boot.initrd.enable = false;
-          system.build.initialRamdisk = pkgs.runCommand "fake-initrd" { } ''
-            mkdir $out
-            touch $out/${config.system.boot.loader.initrdFile}
-          '';
-          system.build.initialRamdiskSecretAppender = pkgs.writeShellScriptBin "append-initrd-secrets" "";
+          # WSL uses its own kernel and boot loader
+          boot = {
+            initrd.enable = false;
+            kernel.enable = false;
+            loader.grub.enable = false;
+            modprobeConfig.enable = false;
+          };
+
+          # WSL does not support virtual consoles
+          console.enable = false;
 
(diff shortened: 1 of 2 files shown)


SuperSandro2000:
https://github.com/NixOS/nixpkgs/pull/197922 didn't make the branch off, backporting

modules/wsl-distro.nix
```diff
@@ -102,20 +103,16 @@ with lib; {
             '';
           };
 
+          # no udev devices can be attached
+          services.udev.enable = false;
```
573, line 107: Could we at least make this overrideable (`lib.mkDefault`) ?
Have the case in mind that someone tries to attach a usb scanner - did that once in a ubuntu wsl instance so should be possible at least.

SuperSandro2000: You can always overwrite it with lib.mkForce.

nzbr: That's true, but I'm in favor of using mkDefault as well. We're disabling udev by default, because it doesn't need to run; disabling it is not necessary for the system to work

nzbr:
@SuperSandro2000 Can you do a flake update? The blocking backport-PR has been merged a while ago

SuperSandro2000:
rebased and updated

nzbr (approved):

PR #177, opened 2022-12-11
Update to NixOS 22.11

nzbr:

PR #178, opened 2022-12-12
Mount propagation fixes (and also oomd)

K900:
ugh.

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -1,4 +1,4 @@
-{ lib, pkgs, config, ... }:
+{ lib, pkgs, config, options, ... }:
 
 with lib; {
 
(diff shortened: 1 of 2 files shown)


K900:
The /dev/shm dance is lifted from https://github.com/arkane-systems/bottle-imp/blob/master/binsrc/imp-executor/__main__.py#L28

modules/wsl-distro.nix
```diff
@@ -177,5 +177,10 @@ with lib; {
             '';
           };
         })
+        # this option doesn't exist on older NixOS, so hack.
+        (lib.optionalAttrs (builtins.hasAttr "oomd" options.systemd) {
```
SuperSandro2000, line 181: Maybe we should just lift the minimal supported version to 22.11

modules/wsl-distro.nix
```diff
@@ -177,5 +177,10 @@ with lib; {
...
+          # OOMd requires cgroup pressure info which WSL doesn't have
```
SuperSandro2000, line 182:
```suggestion
          # OOMD requires cgroup pressure info which WSL doesn't have
```

modules/wsl-distro.nix
K900: That is probably better done in a separate PR.

nzbr: That'd be #177

K900: That's not really dropping support for 22.05 though.

PR #181, opened 2022-12-14
Fix copy-launchers activation script

### LINKED ISSUE #164: rsync: [sender] change_dir failed
## Bug description
On a cold wsl startup:

```
Copying /usr/share/icons
rsync: [sender] change_dir "/nix/store/4c5067qszqva944xynbff1xss8b9z0h2-nixos-system-wsl-22.11.20221110.5f588eb/sw/share/icons" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1327) [sender=3.2.6]
Activation script snippet 'copy-launchers' failed (23)
setting up /etc...
setting up /bin...

[process exited with code 1 (0x00000001)]
```

Reproduced at e9e8d01c5a2f57d330d42c9a772cf67d89b43650

## To Reproduce
```
wsl --shuutdown
(wait 8 seconds)
wsl
```

**Expected behavior**
wsl starts up with no issues

## Logs
See above

nzbr:
Fixes #164

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -88,9 +88,13 @@ with lib; {
             copy-launchers = mkIf cfg.startMenuLaunchers (
               stringAfter [ ] ''
                 for x in applications icons; do
-                  echo "Copying /usr/share/$x"
-                  mkdir -p /usr/share/$x
-                  ${pkgs.rsync}/bin/rsync -ar --delete $systemConfig/sw/share/$x/. /usr/share/$x
+                  echo "setting up /usr/share/''${x}..."
+                  if [[ -d $systemConfig/sw/share/$x ]]; then
+                    mkdir -p /usr/share/$x
+                    ${pkgs.rsync}/bin/rsync -ar --delete $systemConfig/sw/share/$x/. /usr/share/$x
+                  else
+                    rm -rf /usr/share/$x
+                  fi
                 done
               ''
             );


modules/wsl-distro.nix
```diff
@@ -88,9 +88,13 @@ with lib; {
...
                 for x in applications icons; do
-                  echo "Copying /usr/share/$x"
-                  mkdir -p /usr/share/$x
-                  ${pkgs.rsync}/bin/rsync -ar --delete $systemConfig/sw/share/$x/. /usr/share/$x
+                  echo "setting up /usr/share/''${x}..."
+                  if [[ -d $systemConfig/sw/share/$x ]]; then
+                    mkdir -p /usr/share/$x
+                    ${pkgs.rsync}/bin/rsync -ar --delete $systemConfig/sw/share/$x/. /usr/share/$x
+                  else
+                    rm -rf /usr/share/$x
```
SuperSandro2000, line 96: I am not sure if this would help things or make them worse

nzbr: The directory doesn't exist at all on regular NixOS, that's why I altered the script to remove it. I don't know how Windows will react to that though (I mean, it works even if the directories don't exist, the start menu launchers just don't get created then). I haven't been able to find out how to make WSL re-scan for launchers, but in theory, it would be nice if we were able to make it do that as part of the activation script. Do you think leaving the directory in place and just deleting the files inside would be better?

SuperSandro2000: if we want to cleanup the directory we should clean it completely before or delete any file we do not know about. Right now we would miss files usually.

nzbr: The `--delete` option on the rsync command deletes everything in the target directory that doesn't exist at the source

morphykuffour:
Does this fix issue #164? If so can it be merged into the main branch because I am having the same issue.

nzbr:
@morphykuffour I forgot to re-request review. I hop it'll be merged soon

SuperSandro2000 (approved):

PR #182, opened 2022-12-14
Launch commands through a login shell

### LINKED ISSUE #155: Add syschdemd workarounds for login shell behavior
Issue on request from a [discussion on Matrix](https://matrix.to/#/!apXxbtexmNGJaAryzM:nixos.org/$OaIdWGzca_HLAMGGnfyQUOPftgoChQ6h81Nbq_GLc6E?via=nixos.org&via=matrix.org&via=nzbr.de).

## Problem

Launching into NixOS WSL doesn't correctly load user environments because the user shell (`bash` in my case) isn't launched as a login shell. This is also a problem for application launchers exposed through `/usr/share/applications` that are typically referencing their respective executables directly.

Previous work on "login shell" was mentioned with #35 which was closed as part of #56 (although the change introduced `source /etc/set-environment` it does not seem to resolve sourcing `~/.profile`.).

## Solution

The current workaround I'm using is to set the shell command in Windows Terminal to e.g. `bash --login`. A similar workaround can be done with the application launchers by wrapping their commands with a `bash --login -c "<cmd>"`. However, modifications done to the `.desktop` Windows shortcuts are naturally overwritten on subsequent WSL startups.

For native `systemd` this is an issue to be resolved by upstream. However, for `syschdemd` it might be desirable (at least through an option) to wrap commands to ensure they're run in proper user environments.

## Related discussions

https://github.com/nix-community/NixOS-WSL/discussions/128

nzbr:
Fixes #155

### DIFF
--- a/tests/login-shell/session-variables.nix
+++ b/tests/login-shell/session-variables.nix
@@ -1,0 +1,19 @@
+{ pkgs, lib, ... }:
+{
+  imports = [
+    ./base.nix
+    "${builtins.fetchTarball "https://github.com/nix-community/home-manager/archive/master.tar.gz"}/nixos"
+  ];
+
+  home-manager.users.nixos = { ... }: {
+    home = {
+      stateVersion = "22.11";
+      packages = [ pkgs.vim ];
+      sessionVariables = {
+        EDITOR = "vim";
+        TEST_VARIABLE = "THISISATESTSTRING";
+      };
+    };
+    programs.bash.enable = true;
+  };
+}
--- a/tests/login-shell/session-variables.nix
+++ b/tests/login-shell/session-variables.nix
@@ -1,13 +1,16 @@
 { config, pkgs, lib, ... }:
+let
+  ver = with lib; substring 0 5 version;
+in
 {
   imports = [
     ./base.nix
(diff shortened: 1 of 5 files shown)


tests/login-shell/session-variables.nix
```diff
@@ -0,0 +1,19 @@
+{ pkgs, lib, ... }:
+{
+  imports = [
+    ./base.nix
+    "${builtins.fetchTarball "https://github.com/nix-community/home-manager/archive/master.tar.gz"}/nixos"
```
SuperSandro2000, line 5: We shouldn't fetch home-manager from master. This can unexpectedly break.

Also why do we even use home-manager here? We can set session variables with NixOS, too.

nzbr: I used home-manager for the test, because that was specifically what someone had problems with. I'll have to take a look at how sessionVariables work in NixOS

nzbr: I've changed the URL to download the release version corresponding to the current NixOS version. We can't use `environment.sessionVariables`, because, as it turns out, those are set in `/etc/set-environment` which is also sourced outside of a login-shell thus defeating the purpose of the test

PR #183, opened 2022-12-17
nixos-wsl-version command

nzbr:
Add a `nixos-wsl-version` command. This pr also introduces a new versioning scheme to be compatible with the version format that will be required for the Distro Launcher. The format looks as follows:
```
MAJOR.MINOR.TAG.COMMIT
e.g.
22.11.4.78
```
where `MAYOR` and `MINOR` are the NixOS version, `TAG` is incremented with every NixOS-WSL release and `COMMIT` counts the commits since that release. This version is automatically calculated in the CI pipeline. If the module is imported as a flake input or the installer tarball is build locally, the version will show as `DEV_BUILD` instead.
The output of `nixos-wsl-version` contains more information besides the calculated version and looks like this:
```
NixOS-WSL 22.11.4.77 fbe364f47cb713b759865d06e4962b9930516a18 syschdemd
```
That information is additionally available in JSON format
```json
{
  "release": "22.11.4.77",
  "rev": "fbe364f47cb713b759865d06e4962b9930516a18",
  "systemd": "syschdemd"
}
```

`rev` is the git revision of NixOS-WSL and is present in CI builds as well as when importing NixOS-WSL as a flake input. When the repository has been modified locally, it will show `dirty` instead.

`systemd` shows whether native-systemd is enabled and is either `syschdemd` if it isn't or `native` if it is

### DIFF
--- a/modules/version.nix
+++ b/modules/version.nix
@@ -1,0 +1,79 @@
+{ lib, pkgs, options, config, ... }:
+with builtins; with lib;
+{
+
+  options = with types; {
+    wsl.version =
+      let
+        versionFile = splitString "\n" (readFile ../VERSION);
+      in
+      {
+        release = mkOption {
+          description = "the NixOS-WSL release";
+          type = str;
+          default = elemAt versionFile 0;
+        };
+        rev = mkOption {
+          description = "the NixOS-WSL git revision";
+          type = str;
+          default = elemAt versionFile 1;
+        };
+        systemd = mkOption {
+          type = enum [ "native" "syschdemd" ];
+          description = "the systemd implementation used by NixOS-WSL";
+          default = if config.wsl.nativeSystemd then "native" else "syschdemd";
+        };
+      };
+  };
+
+  config = mkIf config.wsl.enable {
+
+    environment.systemPackages = [
+      (
+        with config.wsl.version;
+        let
+          opts = options.wsl.version;
+          maxlen = foldl (acc: opt: max acc (stringLength opt)) 0 ((attrNames opts) ++ [ "help" "json" ]);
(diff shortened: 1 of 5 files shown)


K900 (approved):
LGTM. Maybe we can add more debug info to this later too.

modules/version.nix
```diff
@@ -0,0 +1,79 @@
+{ lib, pkgs, options, config, ... }:
+with builtins; with lib;
```
SuperSandro2000, line 2:
```suggestion
with lib;
```
almost anything under builtins is also available under lib

modules/version.nix
```diff
@@ -0,0 +1,79 @@
...
+          description = "the NixOS-WSL git revision";
+          type = str;
+          default = elemAt versionFile 1;
+        };
+        systemd = mkOption {
+          type = enum [ "native" "syschdemd" ];
+          description = "the systemd implementation used by NixOS-WSL";
+          default = if config.wsl.nativeSystemd then "native" else "syschdemd";
+        };
+      };
```
SuperSandro2000, line 26: Those should be internal.

nzbr: How'd you suggest doing that? I think the values should ideally stay accessible as read-only. `rev` is currently being set in flake.nix

nzbr:
As it turns out, `builtins.toJSON` doesn't appear to be in `lib`

modules/version.nix
SuperSandro2000: I mean this option https://github.com/NixOS/nixpkgs/blob/cc8def18f0afbc28fbdd26d2f5972dc6cb158b7e/nixos/lib/testing/network.nix#L74

nzbr:
@SuperSandro2000 I've now marked the options as internal, like you suggested

PR #203, opened 2023-01-27
Only mask systemd-resolved if WSL is generating resolv.conf

KoviRobi:
I'm using systemd-resolved with tailscale

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -139,7 +139,7 @@ in
             # Disable systemd units that don't make sense on WSL
             services = {
               firewall.enable = false;
-              systemd-resolved.enable = lib.mkDefault false;
+              systemd-resolved.enable = lib.mkDefault (!config.wsl.wslConf.network.generateResolvConf);
               # systemd-timesyncd actually works in WSL and without it the clock can drift
               systemd-timesyncd.unitConfig.ConditionVirtualization = "";
             };


modules/wsl-distro.nix
```diff
@@ -108,19 +108,23 @@ with lib; {
...
+                "autovt@".enable = false;
+                firewall.enable = false;
+                # system clock cannot be changed
+                systemd-timesyncd.enable = false;
+                # no udev devices can be attached
+                systemd-udevd.enable = false;
+              }
+              (mkIf config.wsl.wslConf.network.generateResolvConf {
+                systemd-resolved.enable = false;
+              })
```
SuperSandro2000, line 126:
```suggestion
              systemd-resolved.enable = lib.mkDefault config.wsl.wslConf.network.generateResolvConf;
```
Wouldn't that also work?

nzbr: If it does, we wouldn't even need the mkMerge

KoviRobi: Apart from the logic error of wanting to negate it to `(!config.wsl.wslConf.network.generateResolvConf)` it does work

SuperSandro2000 (approved):

SuperSandro2000:
@KoviRobi can you please rebase?

modules/wsl-distro.nix
```diff
@@ -139,7 +139,7 @@ in
             # Disable systemd units that don't make sense on WSL
             services = {
               firewall.enable = false;
-              systemd-resolved.enable = lib.mkDefault false;
+              systemd-resolved.enable = lib.mkDefault (!config.wsl.wslConf.network.generateResolvConf);
```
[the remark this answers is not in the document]
nzbr: I actually like the `mkDefault (!...)` more, but it's easier to miss the `!` there when reading it in the future, so I don't have a strong opinion about it

K900:
It's `mkDefault`, your explicit setting should override this already.

PR #204, opened 2023-01-27
Default user name isn't necessarily the same as user attr name

KoviRobi:
Me with my weird system again, sorry.

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -104,6 +104,9 @@ with lib; {
               ln -sf ${bash}/bin/sh /bin/sh
               ln -sf ${pkgs.util-linux}/bin/mount /bin/mount
             '';
+            update-entrypoint.text = ''
+              ln -sf ${config.users.users.root.shell} /nix/nixos-wsl/entrypoint
+            '';
           };
 
           systemd = {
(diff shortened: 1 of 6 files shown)


SuperSandro2000 (approved):

nzbr:
@KoviRobi I have opened a PR against yours that adds tests for this https://github.com/KoviRobi/NixOS-WSL/pull/2

SuperSandro2000:
I am actually the same so I am surprised that it works for me and not for you.

modules/wsl-distro.nix
```diff
@@ -153,7 +156,7 @@ with lib; {
         })
         (mkIf cfg.nativeSystemd {
           wsl.wslConf = {
-            user.default = cfg.defaultUser;
+            user.default = config.users.users.${cfg.defaultUser}.name;
```
SuperSandro2000, line 159: ah, probably because my user.users attr matches name.

SuperSandro2000 (approved):

modules/wsl-distro.nix
nzbr: Changing the username has always worked fine, but having different username and attr hasn't

PR #207, opened 2023-01-29
fix: relink the X11 socket into place if needed

K900:
Fixes X11 applications on tmp-as-tmpfs setups.

### DIFF
--- a/modules/wsl-conf.nix
+++ b/modules/wsl-conf.nix
@@ -20,7 +20,7 @@ with lib; {
         description = "Mount entries from /etc/fstab through WSL. You should probably leave this on false, because systemd will mount those for you.";
       };
       root = mkOption {
-        type = str;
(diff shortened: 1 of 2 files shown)


SuperSandro2000:
What is when upstream changes it's tmpfiles.d rule again? Then we have the same problem we had when we removed the rule in https://github.com/nix-community/NixOS-WSL/pull/172

K900:
Looks like upstream is actually just masking the default settings now:

```
~
❯ cat /run/tmpfiles.d/x11.conf
# Note: This file is generated by WSL to prevent systemd-tmpfiles from removing /tmp/.X11-unix during boot.
```

So we should be OK to recreate the link.

nzbr:
@K900 i've opened a PR against your branch. I mount the windows drives to `/drv` instead of `/mnt`, so this didn't work for me. Otherwide it looks good

PR #208, opened 2023-01-30
fix: ensure /nix/nixos-wsl exists before linking entrypoint

K900:


modules/wsl-distro.nix
```diff
@@ -105,6 +105,7 @@ with lib; {
               ln -sf ${pkgs.util-linux}/bin/mount /bin/mount
             '';
             update-entrypoint.text = ''
+              mkdir -p /nix/nixos-wsl
               ln -sf ${config.users.users.root.shell} /nix/nixos-wsl/entrypoint
```
niklaskorz, line 109:
```suggestion
              ln -sfn ${config.users.users.root.shell} /nix/nixos-wsl/entrypoint
```

Not directly related to this pull request, but the `-n` is required here too, otherwise `-f` will only overwrite file links, not directory links.

PR #211, opened 2023-01-30
Add wsl.binShPkg and wsl.populateBin options to make envfs easily pos…

SuperSandro2000:
…sible

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -1,9 +1,27 @@
 { lib, pkgs, config, options, ... }:
 
-with lib; {
+with lib;
+
+let
+  bashWrapper = pkgs.runCommand "nixos-wsl-bash-wrapper"
+    {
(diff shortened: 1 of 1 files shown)


nzbr:
I think instead of using the default, we should set the options in the config section, so that mkForce is required to override them, making it harder to brick the system on accident

modules/wsl-distro.nix
```diff
@@ -14,25 +34,29 @@ with lib; {
...
       description = "The name of the default user";
     };
+    populateBin = mkOption {
+      type = bool;
+      default = true;
+      internal = true;
+      description = ''
+        Dangerous! Things might break. Use with caution!
+
+        Do not populare /bin.
```
niklaskorz, line 44:
```suggestion
        Do not populate /bin.
```

nzbr (approved):

PR #212, opened 2023-02-01
fix: actually fix PATH filtering

K900:


nzbr (approved):
Oof... well, it makes sense at least
(Provided that the tests don't explode)

PR #213, opened 2023-02-01
Add coreutils to bashWrapper

niklaskorz:
VS Code needs the coreutils (to be precise, `dirname` and `uname`) to be available in `/bin/sh` when setting up the remote server.

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -8,7 +8,7 @@ let
       nativeBuildInputs = [ pkgs.makeWrapper ];
     } ''
     makeWrapper ${pkgs.bashInteractive}/bin/sh $out/bin/sh \
-      --prefix PATH ':' ${lib.makeBinPath (with pkgs; [ systemd gnugrep ])}
+      --prefix PATH ':' ${lib.makeBinPath (with pkgs; [ systemd gnugrep coreutils ])}
   '';
 
   cfg = config.wsl;


K900:
In my experience it still works without those?

niklaskorz:
Actually it also complains about `tar` missing in that case...

grafik

niklaskorz:
This is the result when using the current state of the main branch:

grafik

niklaskorz:
Actually when setting up the VS Code WSL remote server from within NixOS with nativeSystemd, it will also error with the `uname` and `dirname` not found. But the `tar` and `gzip` packages are not necessary in that case.

niklaskorz:
As https://github.com/msteen/nixos-vscode-server does not work when setting up the remote server through the GUI (I guess it can't listen to the file changes performed by `/bin/sh`?), I think it's sufficient to only add the coreutils to bashWrapper, and maybe add a note the NixOS-WSL README that nixos-vscode-server is needed to automatically patch the remote server's node binary and only works when initially starting VS Code from within WSL.

modules/wsl-distro.nix
```diff
@@ -29,7 +29,7 @@ with lib; {
       shim = pkgs.callPackage ../scripts/native-systemd-shim/shim.nix { };
 
       bashWrapper = pkgs.runCommand "nixos-wsl-bash-wrapper" { nativeBuildInputs = [ pkgs.makeWrapper ]; } ''
-        makeWrapper ${pkgs.bashInteractive}/bin/sh $out/bin/sh --prefix PATH ':' ${lib.makeBinPath [pkgs.systemd pkgs.gnugrep]}
+        makeWrapper ${pkgs.bashInteractive}/bin/sh $out/bin/sh --prefix PATH ':' ${lib.makeBinPath [pkgs.systemd pkgs.gnugrep pkgs.coreutils]}
```
SuperSandro2000, line 32:
```suggestion
        makeWrapper ${pkgs.bashInteractive}/bin/sh $out/bin/sh \
          --prefix PATH ':' ${lib.makeBinPath (with pkgs; [ systemd gnugrep coreutils ]}
```

niklaskorz:
I'm now simply making use of the new `wsl.binShPkg` option so this is PR is no longer required:

```nix
wsl = pkgs.lib.mkForce {
  binShPkg = with pkgs; runCommand "nixos-wsl-bash-wrapper"
    {
      nativeBuildInputs = [ makeWrapper ];
    } ''
    makeWrapper ${bashInteractive}/bin/sh $out/bin/sh \
      --prefix PATH ':' ${lib.makeBinPath ([ systemd gnugrep coreutils ])}
  '';
};
```

PR #214, opened 2023-02-01
feat: use a Rust tool to do the PATH manipulations

K900:
Overkill? Yes. Fuck bash? Also yes.


nzbr:
I don't know a lot about rust, but this seems fine

modules/wsl-distro.nix
```diff
@@ -179,10 +179,7 @@ with lib; {
             # preserve $PATH from parent
             variables.PATH = [ "$PATH" ];
             extraInit = ''
-              export WSLPATH=$(echo "$PATH" | tr ':' '\0' | command grep -az "^${cfg.wslConf.automount.root}" | tr '\0' ':')
-              ${if cfg.interop.includePath then "" else ''
-                export PATH=$(echo "$PATH" | tr ':' '\0' | command grep -avz "^${cfg.wslConf.automount.root}" | tr '\0' ':')
-              ''}
+              eval $(${nativeUtils}/bin/split-path --automount-root="${cfg.wslConf.automount.root}" ${if cfg.interop.includePath then "--include-interop" else ""})
```
SuperSandro2000, line 182:
```suggestion
              eval $(${nativeUtils}/bin/split-path --automount-root="${cfg.wslConf.automount.root}" ${lib.optionalString cfg.interop.includePath then "--include-interop"})
```

nzbr: The `then` needs to go as well when using `optionalString`

PR #221, opened 2023-02-11
Support CUDA in sandboxes

Atry:
`/usr/lib/wsl` includes CUDA drivers. According to https://discourse.nixos.org/t/pytorch-cuda-on-wsl/18267/2 the drivers can be enabled by `export LD_LIBRARY_PATH=/usr/lib/wsl/lib`. However, without this PR, the approach only works if sandbox is not used.

This PR will enable CUDA in a sandbox, too.

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -26,6 +26,7 @@ in
   nix.package = pkgs.nixFlakes;
   nix.extraOptions = ''
     experimental-features = nix-command flakes
+    extra-sandbox-paths = /usr/lib/wsl
   '';
 
   system.stateVersion = "22.11";


nzbr:
This shouldn't be in `configuartion.nix`, but somewhere in the wsl module, ideally in a new `cuda.nix` file. Furthermore, it shouldn't be enabled by default but through an option, so that the nix sandbox doesn't get modified for people who don't have an Nvidia GPU or don't want to use CUDA

SuperSandro2000:
Closing since this is stale and requires more work

PR #228, opened 2023-03-12
Fix perl locale errrors in installer

TeofilC:
This avoids perl outputting warnings about missing locales during the installer


nzbr:
Thank you!

PR #225, opened 2023-03-01
Enable udev by default (it can be useful when using usbip)

KoviRobi:
We can support usbip with the Microsoft Store version of WSL (I think that's the one that introduced the change, also turned on WSLg)

Related is https://github.com/nix-community/NixOS-WSL/pull/175#discussion_r1048284393 which is where it got disabled, because it wasn't necessary rather than because it broke things

See https://github.com/nix-community/NixOS-WSL/issues/111 for NixOS-WSL usbip discussion

PR #232, opened 2023-03-15
Revert "Enable udev by default (it can be useful when using usbip)"

SuperSandro2000:
Reverts nix-community/NixOS-WSL#225

Going to revert this because it draws in firmware which is 300MB big and cannot easily be disabled.

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -127,6 +127,9 @@ in
             '';
           };
 
+          # no udev devices can be attached
+          services.udev.enable = lib.mkDefault false;
+
           systemd = {
             # Disable systemd units that don't make sense on WSL
             services = {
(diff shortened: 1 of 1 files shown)


modules/wsl-distro.nix
```diff
@@ -127,6 +127,9 @@ in
             '';
           };
 
+          # no udev devices can be attached
```
SuperSandro2000, line 130:
```suggestion
          # useful for usbip but adds a dependency on various firmwares which are combined over 300 MB big
```

KoviRobi:
Totally fair, I wonder what I did wrong when I compared sizes in the tarball

nzbr:
We could do
```
  hardware.firmware = mkOverride 60 [ ];
```
but I guess usbip will need the firmware files

SuperSandro2000:
Also reversing that is in question much harder than just reenabling udev.

PR #233, opened 2023-03-26
wsl-conf: warn when networking.extraHosts has no effect

### LINKED ISSUE #230: networking.extraHosts doesn't work
## Bug description
Just using `networking.extraHosts` will not change the `/etc/hosts` file since that one is by default managed by WSL. 

## To Reproduce
set
`config.networking.extraHosts = 'test 127.0.0.1'` 

**Expected behavior**
I think either there should be an assertion that makes sure that `wsl.wslConf.network.generateHosts` and `networking.extraHosts` aren't set together, or even more ideally would be ofc a solution where we could still append to the wsl hosts config ^^

SuperSandro2000:
fixes #230

### DIFF
--- a/modules/wsl-conf.nix
+++ b/modules/wsl-conf.nix
@@ -85,9 +85,10 @@ with lib; {
 
     environment.etc."wsl.conf".text = generators.toINI { } config.wsl.wslConf;
 
-    warnings = (optional (config.wsl.wslConf.boot.systemd && !config.wsl.nativeSystemd)
-      "systemd is enabled in wsl.conf, but wsl.nativeSystemd is not enabled. Unless you did this on purpos, this WILL make your system UNBOOTABLE!"
-    );
+    warnings = optional (config.wsl.wslConf.boot.systemd && !config.wsl.nativeSystemd)
+      "systemd is enabled in wsl.conf, but wsl.nativeSystemd is not enabled. Unless you did this on purpose, this WILL make your system UNBOOTABLE!"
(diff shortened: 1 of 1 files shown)


modules/wsl-conf.nix
```diff
@@ -85,9 +85,10 @@ with lib; {
 
     environment.etc."wsl.conf".text = generators.toINI { } config.wsl.wslConf;
 
-    warnings = (optional (config.wsl.wslConf.boot.systemd && !config.wsl.nativeSystemd)
+    warnings = optional (config.wsl.wslConf.boot.systemd && !config.wsl.nativeSystemd)
       "systemd is enabled in wsl.conf, but wsl.nativeSystemd is not enabled. Unless you did this on purpos, this WILL make your system UNBOOTABLE!"
```
nzbr, line 89:
```suggestion
      "systemd is enabled in wsl.conf, but wsl.nativeSystemd is not enabled. Unless you did this on purpose, this WILL make your system UNBOOTABLE!"
```

When you're editing this file anyway, can you fix that typo as well?

modules/wsl-conf.nix
```diff
@@ -85,9 +85,10 @@ with lib; {
...
-    );
+    ++ lib.optional (config.wsl.wslConf.network.generateHosts && config.networking.extraHosts != "")
```
nzbr, line 90:
```suggestion
    ++ optional (config.wsl.wslConf.network.generateHosts && config.networking.extraHosts != "")
```

nzbr (changes requested):
Looks good otherwise

nzbr (approved):

PR #240, opened 2023-04-04
fix: explicitly add systemctl and grep to path in shell-wrapper

K900:
We can generally find them in /run/current-system/sw/bin, but WSL may try to call those before activation is fully done, so make sure it can always find what it needs.

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -4,6 +4,7 @@ with lib;
 
 let
   bashWrapper = pkgs.writeShellScriptBin "sh" ''
+    export PATH=${lib.makeBinPath [ pkgs.systemd pkgs.gnugrep ]}
     . ${config.system.build.etc}/etc/set-environment
     exec ${pkgs.bashInteractive}/bin/sh "$@"
   '';


modules/wsl-distro.nix
```diff
@@ -4,6 +4,7 @@ with lib;
 
 let
   bashWrapper = pkgs.writeShellScriptBin "sh" ''
+    export PATH=${lib.makeBinPath [ pkgs.systemd pkgs.gnugrep ]}
     . ${config.system.build.etc}/etc/set-environment
```
SuperSandro2000, line 8: Doesn't this PATH get set here? Not sure where a command should be executed here. I am probably missing some context

modules/wsl-distro.nix
```diff
@@ -4,6 +4,7 @@ with lib;
 
 let
   bashWrapper = pkgs.writeShellScriptBin "sh" ''
+    export PATH=${lib.makeBinPath [ pkgs.systemd pkgs.gnugrep ]}
```
SuperSandro2000, line 7:
```suggestion
    export PATH=${lib.makeBinPath with pkgs; [ systemd gnugrep ]}
```

modules/wsl-distro.nix
SuperSandro2000: hmmm, I see. We should comment that.

nzbr:
If we're doing that, shouldn't we just add `environment.systemPackages` (and `environment.defaultPackages`) to path instead of some packages explicitly and sourcing `/etc/set-environment`?

K900:
I'm not sure, I'd rather have the bare minimum needed by the WSL machinery here and handle the rest in a way that's as conventional as possible.

nzbr:
We could also just put a busybox in there. That would be future-proof as well and less easy to brick by changing the system config

K900:
They do call systemctl, so we can't _just_ put a busybox in there.

nzbr:
Right, I meant that more as "instead of `gnugrep`"

K900:
I'm 100% expecting WSL to assume gnugrep there tbh.

nzbr (approved):

PR #242, opened 2023-04-11
enable xvfb and other X servers

jcumming:
Currently, /tmp/.X11-unix is a ro bind mount, this makes starting other X servers (Xvfb, vnc, etc) impossible. This patch unmounts the bind mount, makes the directory writable, and creates a symlink to the wslg passthrough socket.


modules/wsl-distro.nix
```diff
@@ -112,6 +112,11 @@ in
                 done
               ''
             );
+            umountTmpX11.text = ''
```
SuperSandro2000, line 115:
```suggestion
            umountTmpX11 = ''
```

modules/wsl-distro.nix
```diff
@@ -112,6 +112,11 @@ in
...
+              # this will be created again rw by systemd.tmpfiles.rules
+              echo "unmounting readonly /tmp/.X11-unix..."
```
SuperSandro2000, line 117:
```suggestion
              echo "unmounting tmpfs at /tmp/.X11-unix..."
```

PR #253, opened 2023-06-14
Enable systemd-timesyncd to prevent clock drift

SuperSandro2000:


nzbr:
Have you tested if this breaks if there is no timezone set? I could imagine it setting the clock to UTC then

SuperSandro2000:
Yes, that will happen https://search.nixos.org/options?channel=unstable&show=time.timeZone&from=0&size=50&sort=relevance&type=packages&query=time.timeZone

PR #260, opened 2023-06-19
Allow customizing the system config to include in the tarball

terlar:
If you build your own custom tarballs this is useful.

### DIFF
--- a/modules/build-tarball.nix
+++ b/modules/build-tarball.nix
@@ -51,17 +51,20 @@ let
     # Write wsl.conf so that it is present when NixOS is started for the first time
     cp ${config.environment.etc."wsl.conf".source} ./etc/wsl.conf
 
-  '' + lib.optionalString cfg.includeConfig (if cfg.configPath == null then ''
-    # Copy the system configuration
-    mkdir -p ./etc/nixos/nixos-wsl
-    cp -R ${lib.cleanSource ../.}/. ./etc/nixos/nixos-wsl
-    mv ./etc/nixos/nixos-wsl/configuration.nix ./etc/nixos/configuration.nix
-    # Patch the import path to avoid having a flake.nix in /etc/nixos
-    sed -i 's|import \./default\.nix|import \./nixos-wsl|' ./etc/nixos/configuration.nix
-  '' else ''
-    mkdir -p ./etc/nixos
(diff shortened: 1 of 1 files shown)


terlar:
@nzbr I hope the ping is not disturbing, just wanted to make sure this one was not missed, seems it requires approval to run the workflows as well. But I have tested this locally and it works.

I don't think it is a controversial change, however, I was slightly unsure about the option name.

modules/build-tarball.nix
```diff
@@ -49,14 +49,17 @@ let
...
       cp -R ${lib.cleanSource ../.}/. ./etc/nixos/nixos-wsl
       mv ./etc/nixos/nixos-wsl/configuration.nix ./etc/nixos/configuration.nix
       # Patch the import path to avoid having a flake.nix in /etc/nixos
       sed -i 's|import \./default\.nix|import \./nixos-wsl|' ./etc/nixos/configuration.nix
-    ''}
+    '' else ''
+      mkdir -p ./etc/nixos
+      cp -R ${lib.cleanSource config.wsl.tarball.configPath}/. ./etc/nixos
+    '')}
   '';
```
SuperSandro2000, line 63:
```suggestion
    '');
```
didn't test but instead of inlining the string my idea is to append it to reduce the nested things here

terlar:
Not sure if this is cleaner or not, I had to add wrapping parenthesis around the whole string to get it to work

SuperSandro2000 (approved):

terlar:
Would it make sense to chown these files to the defaultUser and add write permissions? I just tested this and in order to modify these after the install I first had to do that, or edit the files via sudo and force write, since it didn't have write permissions configured.

What do you think?

SuperSandro2000:
This is the standard nixos configuration and I am not sure if we should change that.

nzbr:
I don't think the default user should be able to write the config, but giving root write permissions on the files would definitely be a good idea (and something I wanted to make an issue for multiple times, but always forgot)

terlar:
I pushed another commit, giving the root write permissions to the `/etc/nixos` files. I opted for string interpolation again for this section to avoid parenthesis overload. Let me know if you prefer another style.

nzbr (approved):

PR #263, opened 2023-06-20
Add usbip module for usb support in WSL

terlar:
This adds a module to enable USB support via usbip together with a [usbip-win](https://github.com/dorssel/usbipd-win) running on the host.

- Adds the usbip package to be able to run adhoc usbip command e.g. (usbip list --remote 172.23.96.1)
- Enables udev
- Creates systemd services to auto-attach USB devices based on busid

One caveat with accessing USB devices is that they will require wider permissions than that will most likely be configured. For example the infrastructure provided by NixOS to enable YubiKeys will add a rule with permission mode `0660`. While when running WSL with usbip you need to set `0666` in the udev rules.

So for example to get YubiKeys working you would need something like this:
```nix
{
  services = {
    pcscd.enable = true;
    udev = {
      packages = [pkgs.yubikey-personalization];
      extraRules = lib.optionalString config.wsl.usbip.enable ''
        SUBSYSTEM=="usb", ATTRS{idVendor}=="1050", ATTRS{idProduct}=="0010|0110|0111|0114|0116|0401|0403|0405|0407|0410", MODE="0666"
      '';
    };
  };
}
```

If we look into the udev rule provided by the `yubikey-personalization` package it looks like this:
```
$ nix build nixpkgs#yubikey-personalization
$ cat result/lib/udev/rules.d/69-yubikey.rules
ACTION!="add|change", GOTO="yubico_end"

# Udev rules for letting the console user access the Yubikey USB
# device node, needed for challenge/response to work correctly.

# Yubico Yubikey II
ATTRS{idVendor}=="1050", ATTRS{idProduct}=="0010|0110|0111|0114|0116|0401|0403|0405|0407|0410", \
    ENV{ID_SECURITY_TOKEN}="1"

LABEL="yubico_end"
```

I don't know enough about udev rules, but would it be possible to do something generic like this:
```
SUBSYSTEM=="usb", MODE="0666"
```

Any ideas appreciated. At least with this configuration I have YubiKey's working within WSL with auto-attach, so I just plug in/unplug and can use it like I would regularly expect on a normal Linux device.

### DIFF
--- a/modules/usbip.nix
+++ b/modules/usbip.nix
@@ -1,0 +1,51 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  usbipd-win-auto-attach = pkgs.fetchurl {
+    url = "https://raw.githubusercontent.com/dorssel/usbipd-win/v3.1.0/Usbipd/wsl-scripts/auto-attach.sh";
+    hash = "sha256-KJ0tEuY+hDJbBQtJj8nSNk17FHqdpDWTpy9/DLqUFaM=";
+  };
+
+  cfg = config.wsl.usbip;
+in
+{
+  options.wsl.usbip = with types; {
+    enable = mkEnableOption "USB/IP integration";
+    autoAttach = mkOption {
+      type = listOf str;
+      default = [ ];
+      example = [ "4-1" ];
+      description = "Auto attach devices with provided Bus IDs.";
+    };
+  };
+
+  config = mkIf (config.wsl.enable && cfg.enable) {
+    environment.systemPackages = [
+      pkgs.linuxPackages.usbip
+    ];
+
+    services.udev.enable = true;
+
+    systemd = {
+      services."usbip-auto-attach@" = {
+        description = "Auto attach device having busid %i with usbip";
+        after = [ "network.target" ];
+
+        scriptArgs = "%i";
+        path = [ pkgs.linuxPackages.usbip ];
+
+        script = ''
+          busid="$1"
+          ip="$(grep nameserver /etc/resolv.conf | cut -d' ' -f2)"
+
+          echo "Starting auto attach for busid $busid on $ip."
+          source ${usbipd-win-auto-attach} "$ip" "$busid"
+        '';
+      };
+
+      targets.multi-user.wants = map (busid: "usbip-auto-attach@${busid}.service") cfg.autoAttach;
+    };
+  };
+}
--- a/flake.nix
+++ b/flake.nix
@@ -19,7 +19,7 @@
         imports = [
           ./modules
 
-          ({ ... }: {
+          (_: {
             wsl.version.rev = mkIf (self ? rev) self.rev;
           })
         ];
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -5,6 +5,7 @@
     ./interop.nix
     ./recovery.nix
     ./systemd
+    ./usbip.nix
     ./version.nix
     ./welcome.nix
     ./wsl-conf.nix


modules/usbip.nix
```diff
@@ -0,0 +1,58 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  usbipd-win-auto-attach = pkgs.fetchurl {
+    url = "https://raw.githubusercontent.com/dorssel/usbipd-win/9e259e74e225ed9321484366a79545b00a18185c/Usbipd/wsl-scripts/auto-attach.sh";
```
SuperSandro2000, line 7: Can we use the release tags?

modules/usbip.nix
```diff
@@ -0,0 +1,58 @@
...
+    enable = mkEnableOption "USB/IP integration";
+    autoAttach = mkOption {
+      type = listOf str;
+      default = [ ];
+      example = [ "4-1" ];
+      description = "Auto attach devices with provided Bus IDs.";
+    };
+  };
+
+  config = mkIf (config.wsl.enable && cfg.enable) {
```
SuperSandro2000, line 24:
```suggestion
  config = mkIf cfg.enable {
```
or do we have similar checks on other modules?

modules/usbip.nix
```diff
@@ -0,0 +1,58 @@
...
+
```
SuperSandro2000, line 25:
```suggestion
```

modules/usbip.nix
```diff
@@ -0,0 +1,58 @@
...
+    systemd = {
+      services."usbip-auto-attach@" = {
+        description = "Auto attach device having busid %i with usbip";
+        after = [ "network.target" ];
+
+        scriptArgs = "%i";
+        path = [
+          pkgs.coreutils
+          pkgs.gnugrep
```
SuperSandro2000, line 40: Aren't those there by default?

modules/usbip.nix
terlar: Yeah, that is where I got it, I'm not sure how likely it is that someone configures the sub option but not the top one. But it could allow for some kind of composition/partial configuration. Thought it had a potential use case. But I'm fine either way

modules/usbip.nix
terlar: You might be correct. I'll try to remove them and see if the script executes.

nzbr:
Why do devices need to have `0666` permissions? They don't need them on bare-metal NixOS, so why would they need them here? Shouldn't it be possible to add the user to the appropriate groups instead?

terlar:
Okay, this is a bit strange, so I investigated a bit how it works on a pure NixOS machine as opposed to NixOS in WSL.

On WSL my USB-device gets:
```
crw-rw---- 1 root root 189,   2 Jun 26 02:24 /dev/bus/usb/001/003
```

However on NixOS for the same device I get:
```
crw-rw-rw-+ 1 root root 189,   2 Jun 26 02:24 /dev/bus/usb/001/018
```

So somehow the pure NixOS version applied some udev-rule to make that device writeable.

On pure NixOS it also appeared under `/dev/input` which didn't exist at all with WSL.

Another option I checked was that I can assign a group like so:
```
SUBSYSTEM=="usb", GROUP="usb"
```

But then how it works would depart. I will investigate a bit more to see if I can find out why the udev rules evaluate differently.

terlar:
On bare-metal NixOS it has the "same" permissions, but it has a ACL associated with the usb device.
```
$ ls -la /dev/bus/usb/001/007
crw-rw----+ 1 root root 189,   2 Jun 26 02:24 /dev/bus/usb/001/007
$ getfacl /dev/bus/usb/001/007
# file: dev/bus/usb/001/007
# owner: root
# group: root
user::rw-
user:myuser:rw-
group::rw-
mask::rw-
other::r--
```

I'm new to the ACL:s and just found out about them now. But that seems to be the difference.

modules/usbip.nix
terlar: A release have not happened, so I can use the release tag.

terlar:
@nzbr and @SuperSandro2000, any ideas how to proceed? The script has now received a tag, so I have updated that. Other than that I guess it is a few open questions. I also noticed a test failed, but I am not sure if it is related to my changes?

SuperSandro2000:
The test failure is related to docker being broken on the commit on stable.

SuperSandro2000 (approved):

nzbr:
@terlar Have you figured something out regarding udev? If there's no other easy way, giving all USB devices `0666` sounds like the best solution

terlar:
@nzbr No unfortunately not, only the findings I mentioned recently. So the permissions are the same on NixOS standalone and NixOS in WSL. But NixOS standalone has ACL configured for the user so the user gets access. Not sure what is configuring the ACL and why they wouldn't be configured inside WSL.

Either we leave it to the users or give the 0666 permissions to all USB devices. Up to you, I'm fine either way. I guess you are not familiar with the ACL stuff either?

SuperSandro2000 (approved):
.

terlar:
Seems there was another conflict due to recent changes in `modules/default.nix`. I rebased on the main branch again.

I'm okay to go either way, either like this is or with modifications. Just let me know what to do :)

LGUG2Z:
FWIW I've been using the additions on this branch with the suggested `udev` rule without any problems for a week or so.

For people who go a step further and build a custom WSL kernel with HIDRAW enabled to by able to use FIDO2-related features, this is the extra `udev` rule I added to allow me to use the FIDO2-related `ykman` commands (+FIDO2 SSH) without `sudo`:

```
KERNEL=="hidraw*", SUBSYSTEM=="hidraw", TAG+="uaccess", MODE="0666"
```

573:
What is the state of affairs for this ?

terlar:
I guess it is waiting for decision on proceed or if there is a way to not have those udev rules or perhaps add them by default.

573:
For me the question was just how to make use of all of this from the user side as depite all of the guides it was still not entirely made clear how to attach usb devices.

Right I'm using @LGUG2Z 's sumup to get a working module on the nixos side with the udev rules @terlar provided in the PR, still the following intel was missing for me which give still not a full-blown readme style explaination:

At least it seems to be a one-time needed thing: `usbipd bind --busid=1-1 -f` (on further calls with different devices I left the `-f` switch off and it still worked) in admin powershell. If I don't do that, the `usbip list` (in WSL VM) call below gets to answer `usbip: info: no exportable devices found on <ip>`. I can prove in any powershell session (even non-admin) that it worked by saying `usbipd list`: Next to the device in question something like `Attached` (to be fair in my case again I said `usbipd wsl list` and it answered `Attached - non-WSL` then, cool so it is possible to bind devices without providing an instance of WSL) should appear.

On NixOS side: `usbip list --remote="$(grep nameserver /etc/resolv.conf | cut -d' ' -f2)"` (see usb-win's `auto-attach` / @terlar 's PR) is the magic incantation that gives all I need to proceed for now.

Still not clear for me is why my usb scanner isn't `scanimage -L` listed but that belongs to another discussion frankly.

Issue #237, opened 2023-03-31
Missing wslpath in PATH on syschdemd


celestefox:
## Bug description

Having just updated NixOS-WSL for the first time in a while and built a new generation, there is no `wslpath` in my `$PATH`, running with syschdemd still instead of native systemd. I have a `/bin/wslpath` that works, but it is not in my `$PATH`. This is because #209 removed the wrapper from `systemPackages`, and I believe that this still works under native systemd, due to https://github.com/nix-community/NixOS-WSL/blob/main/modules/wsl-distro.nix#L199 inheriting all of `$PATH`, which syschdemd does not do, especially as `wsl -e /usr/bin/env` reports `PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/usr/lib/wsl/lib:` before my Windows paths, which includes `/bin`.

## To Reproduce

Build an installer from `main`, install as a new temporary distro, as the configuration of the installer does not use native systemd, see that there is no `wslpath` command if you try to run it, but `/bin/wslpath` exists and works.

**Expected behavior**

A working `wslpath` on syschdemd.

## Logs
```
wsl -d nixostmp
                      Unpacking root file system...
 366MiB 0:00:15 [23.0MiB/s] [============================================================================================================================================================================================>] 100%
Activating nix configuration...
setting up /usr/share/applications...
setting up /usr/share/icons...
setting up /etc...
setting up /bin...
Cleaning up installer files...
finding garbage collector roots...
removing stale link from '/nix/var/nix/gcroots/auto/6yfbrv8wfkyl40wq6827lamr9m29z7wg' to '/build/nix/var/nix/profiles/per-user/root/channels-1-link'
removing stale link from '/nix/var/nix/gcroots/auto/zxi8v969la5n36bg7gs016wgcif4qjhp' to '/build/nix/var/nix/profiles/system-1-link'
deleting garbage...
deleting '/nix/store/vzdmnirlszj446barmaiyv1mkla89lln-wsl-prepare'
deleting '/nix/store/qk5j2w5901d2qlp535a0z3acm6fmg3p6-tarball'
deleting '/nix/store/96vxkz49y5cnnkzfgln1nksbh0kmkwsz-installer.sh'
deleting '/nix/store/khqypqm75yd71p9xn143r528jf12r7zw-source'
[...]

PR #267, opened 2023-07-12
fix: getent missing in syschdemd

nzbr:
Related to the discussion in #237 (though it doesn't address the original issue, so don't close that after merging this)

`getent` was moved to its own derivation and removed from `glibc` on unstable. Therefore syschdemd crashed on systems running nixos-unstable. By using the getent derivation instead, it works on 22.11, 23.05 and unstable (that's the ones I tested, it might work on older versions as well)

### DIFF
--- a/scripts/syschdemd.nix
+++ b/scripts/syschdemd.nix
@@ -3,7 +3,7 @@
 , lib
 , coreutils
 , daemonize
-, glibc
+, getent
 , gnugrep
 , systemd
 , util-linux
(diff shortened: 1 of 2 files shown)


SuperSandro2000 (approved):

scripts/syschdemd.nix
```diff
@@ -43,7 +43,7 @@ mkWrappedScript {
     "/run/wrappers" # mount
     coreutils
     daemonize
-    glibc # getent
+    getent
```
SuperSandro2000, line 46: we could also use glibc.getent but using getent is more cross friendly even if that does not matter here https://github.com/NixOS/nixpkgs/blob/3a671cd64d03f4e55eb6c817d6294910e0ccc133/pkgs/top-level/unixtools.nix#L74-L78

PR #274, opened 2023-08-09
feat: add a recovery shell

K900:


README.md
```diff
@@ -40,6 +40,26 @@ If you want to make NixOS your default distribution, you can do so with
...
+
+```sh
+wsl -d NixOS --system --user root -- /mnt/wslg/distro/bin/nixos-wsl-recovery
+```
+
+This will load the WSL "system" distribution, activate your configuration,
+then chroot into your NixOS system, similar to what `nixos-enter` would do
+on a normal NixOS install.
+
+You can choose an older generation to load with
```
SuperSandro2000, line 55: Should we add a subcommand to list available generations?

modules/wsl-distro.nix
```diff
@@ -8,6 +8,16 @@ let
     exec ${pkgs.bashInteractive}/bin/sh "$@"
   '';
 
+  nixos-enter' = config.system.build.nixos-enter.overrideAttrs (_: {
```
SuperSandro2000, line 11: Do we already have the fix for https://github.com/NixOS/nix/issues/7602 ?

README.md
K900: I don't really want to make that wrapper any bigger than it needs to be, but maybe we should document that.

modules/wsl-distro.nix
K900: It should be in the next 2.13 release.

K900:
I don't think we should have things in `/nix` that are not managed by Nix itself. It's just asking for trouble later. Otherwise agreed.

README.md
K900: Once we have it, yes, though I'd like to also have the fallback for people that install from their own tarballs.

nzbr:
Good point. How about `/usr/lib/nixos-wsl` then?

K900:
I'm not sure we want to be touching /usr either... Not really sure what a good place for this is.

nzbr:
image
`/usr` and `/usr/lib` look like this on my installation. I don't see how putting some extra files in there could break anything. 
A great place (but I don't know how feasible it is) would be in the toplevel derivation. That would make the binaries independent of the currently active system so that it'd be possible to just use the one from the previous generation when something breaks the recovery script

K900:
Also, I've changed the shebang and added an explicit error message to detect when the recovery shell is running from inside NixOS, so we can abort.

nzbr:
> we can't actually put this in the toplevel, because the toplevel is an absolute symlink

Unfortunate... It might be worth revisiting when adding a recovery command to the launcher, but for now I'll merge this as-is

Issue #277, opened 2023-08-16
Drop support for EOL versions of WSL


K900:
The Windows Store version of WSL is [available for everyone running Windows 10 21H1 or up](https://devblogs.microsoft.com/commandline/the-windows-subsystem-for-linux-in-the-microsoft-store-is-now-generally-available-on-windows-10-and-11/).

Windows 10 21H1 is [EOL since December](https://learn.microsoft.com/en-us/lifecycle/products/windows-10-home-and-pro).

We should remove support for those versions, which will _dramatically_ simplify maintenance.

Things that can be removed once we only support Store WSL:
- [the installer](https://github.com/nix-community/NixOS-WSL/blob/main/modules/installer.nix) can be removed, [tarball generation](https://github.com/nix-community/NixOS-WSL/blob/main/modules/build-tarball.nix) substantially rewritten 
  - the purpose of the installer was to workaround a libarchive bug with `wsl --import` which is resolved on Store
  - being addressed in #243
- [Docker native workarounds](https://github.com/nix-community/NixOS-WSL/blob/main/modules/docker-native.nix) can be fully removed, the latest Store kernel supports nftables
- [interpreter magic](https://github.com/nix-community/NixOS-WSL/blob/main/modules/interop.nix) can be removed in favor of a consistent registration
- [old tmpfiles workarounds](https://github.com/nix-community/NixOS-WSL/blob/main/modules/wsl-distro.nix#L152) can be dropped (probably more?)
- non-native-systemd code can be dropped
  - [all](https://github.com/nix-community/NixOS-WSL/blob/main/modules/wsl-distro.nix#L176-L191)
  - [of](https://github.com/nix-community/NixOS-WSL/blob/main/scripts/syschdemd.nix)
  - [it](https://github.com/nix-community/NixOS-WSL/blob/main/scripts/syschdemd.sh)
  - [gone](https://github.com/nix-community/NixOS-WSL/blob/main/scripts/wrapper.sh)
- maybe more?

K900:
Also, pretty much any of those items can be done independently, so PRs are appreciated.

K900:
Mostly done for 23.05, still need to kill syschdemd.

PR #280, opened 2023-08-26
WIP: implement #277

K900:
Very wip, just to get an idea of diff size.


K900:
Closing until we figure out the tests situation.

PR #281, opened 2023-08-27
feat: use internal WSL knobs instead of symlinks

K900:
This makes things so much easier and more consistent and it's undocumented API that I found in Ghidra so nothing could possibly go wrong.

### DIFF
--- a/modules/wsl-conf.nix
+++ b/modules/wsl-conf.nix
@@ -41,6 +41,21 @@ with lib; {
         default = false;
         description = "Use systemd as init. There's no need to enable this manually, use the wsl.nativeSystemd option instead";
       };
+      initPath = mkOption {
+        type = path;
+        default = null;
+        internal = true;
+      };
+      initShutdownCommand = mkOption {
+        type = path;
+        default = null;
+        internal = true;
+      };
+      initWaitCommand = mkOption {
+        type = path;
+        default = null;
+        internal = true;
+      };
     };
     interop = {
       enabled = mkOption {
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -3,11 +3,6 @@
 with lib;
 
 let
-  bashWrapper = pkgs.writeShellScriptBin "sh" ''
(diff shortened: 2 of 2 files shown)


K900:
Filed an issue with upstream to maybe stabilize these: https://github.com/microsoft/WSL/issues/10425

nzbr:
This looks so much better. I hope they decide to keep these

modules/wsl-conf.nix
```diff
@@ -41,6 +41,21 @@ with lib; {
         default = false;
         description = "Use systemd as init. There's no need to enable this manually, use the wsl.nativeSystemd option instead";
       };
+      initPath = mkOption {
+        type = path;
```
SuperSandro2000, line 45: Doesn't this need to be ``nullOr path``?

modules/wsl-distro.nix
```diff
@@ -211,15 +206,20 @@ in
         (mkIf cfg.nativeSystemd {
           wsl.wslConf = {
             user.default = config.users.users.${cfg.defaultUser}.name;
-            boot.systemd = true;
+            boot = {
+              systemd = true;
+              initPath = "${nativeUtils}/bin/systemd-shim";
+              initShutdownCommand = pkgs.writeShellScript "nixos-wsl-init-shutdown" ''
+                ${pkgs.systemd}/bin/systemctl reboot
```
SuperSandro2000, line 213:
```suggestion
                ${pkgs.systemd}/bin/systemctl reboot "$@"
```
just incase

modules/wsl-conf.nix
K900: Probably, I haven't actually tested this on non-native-systemd yet.

modules/wsl-distro.nix
K900: I'm actually pretty sure we can avoid the shell script here and just write the string directly.

K900:
This is drifting out of sync and I don't want to maintain it until upstream decides this is OK to do. Closing for now, we can revisit later.

Issue #279, opened 2023-08-25
boot.initrd.systemd.extraBin doesn't seem to work


alexvorobiev:
## Bug description
While trying to get `usbip` to work I found that the commands that need to be run from Windows shell can't find `cat` and `usbip` on NixOS (details here https://github.com/nix-community/NixOS-WSL/issues/111#issuecomment-1686372656) but everything works if those two commands are symlinked to either `/bin` or `/usr/bin`. I tried using `boot.initrd.systemd.extraBin` to create the symlinks automatically but that didn't seem to work. Am I using this option incorrectly or is there any other way to do that?

## To Reproduce
Use `boot.initrd.systemd.extraBin` NixOS option to add symlinks to `bin`.

**Expected behavior**
`boot.initrd.systemd.extraBin` adds symlinks to `/bin`.

nzbr:
We do not control the kernel and initrd, because they are a part of WSL. Because of that, this option has no effect. For the specific use case of usbip, we'll probably add an option like `wsl.usbip.enable`.

For everything else, it'd probably make sense to add some way of adding additional things to `/bin`, but we don't have that as of now

PR #291, opened 2023-09-11
Refactor

nzbr:
I separated some code into its own modules mainly to put everything related to syschdemd in one spot so we can remove it later. This also adds a new `wsl.extraBin` option to facilitate that and address #279


K900 (approved):
LGTM at a glance.

PR #295, opened 2023-09-29
New release preparation

K900:
This should hopefully unblock us, finally.


nzbr:
Why does it create a shell script instead of a tarball? Can't we just run whatever that script does inside a derivation?

K900:
No, we need to run the script as root and not in the sandbox so we can do chroot stuff.

modules/interop.nix
```diff
@@ -18,18 +14,6 @@ with builtins; with lib;
       default = true;
       description = "Include Windows PATH in WSL PATH";
     };
-
-    preserveArgvZero = mkOption {
```
SuperSandro2000, line 22: Should we add a mkRemovedOption?

SuperSandro2000 (approved):
I didn't test this in depth but big thanks for tackling this.

modules/interop.nix
K900: It's an internal option so I don't know.

nzbr (approved):
Just tested the builder locally (with nix run github:...) and it seems to work flawlessly

terlar:
Thank you for a lot of fixes and finally having installer-free environment. It is a bit unfortunate that one have to use sudo/root, but I guess that is the case if chroot is needed...

One regression I discovered is the inability to configure the stuff that ends up in `/etc/nixos/...`. Having the ability to place custom code there allows for packing a tarball with much nicer experience.

Now the out of the box experience is:
```
       error: file 'nixos-wsl/modules' was not found in the Nix search path (add it using $NIX_PATH or -I)

       at «none»:0: (source not available)

[user@wsl:~]$ sudo nixos-rebuild -I nixos-config=/etc/nixos/configuration.nix switch
```

I find it nice for the user to be able to iterate on a base and so in my own packaging of the tarball I bundled at flake that pulls in the appropriate nixos-wsl module etc.

Where should such config live? I can make a PR to re-introduce this feature.

K900:
That error isn't about custom code, it's just about channels not being included in the tarball. We can probably run `nix-channel --update` at tarball build time. As for custom config, maybe we should just have an option for adding arbitrary files to the tarball...

PR #303, opened 2023-10-02
feat: add welcome message

K900:
Direct people to upgrade their systems first and link to support channels just in case.


nzbr:
Sounds like a good idea. This is probably something we want to move to the launcher once that's done (it could just print the message after the tarball has been imported)

PR #304, opened 2023-10-02
Bring back customizing the system config to include in the tarball

terlar:
This was previously implemented in #260, but lost in the migration off the installer tarball.


modules/build-tarball.nix
```diff
@@ -34,13 +34,22 @@ let
...
+
   # These options make no sense without the wsl-distro module anyway
-  config = mkIf cfg.enable {
+  config = mkIf config.wsl.enable {
     system.build.tarballBuilder = pkgs.writeShellApplication {
       name = "nixos-wsl-tarball-builder";
 
       runtimeInputs = [
         pkgs.coreutils
+        pkgs.findutils
```
K900, line 52: Doesn't seem used anymore?

modules/default.nix
```diff
@@ -14,7 +14,5 @@
       "Additional workarounds are no longer required for Docker to work. Please use the standard `virtualisation.docker` NixOS options.")
     (lib.mkRemovedOptionModule [ "wsl" "interop" "preserveArgvZero" ]
       "wsl.interop.preserveArgvZero is now always enabled, as used by modern WSL versions.")
-    (lib.mkRemovedOptionModule [ "wsl" "tarball" ]
```
K900, line 17: This needs to be adjusted, not removed, because there were other options in this namespace.

K900:
Generally, I'd probably prefer to have a more generic mechanism of just adding arbitrary files, similar to what `make-system-tarball.nix` does. We can then use that mechanism to include the default config _and_ provide user-level overrides.

terlar:
I updated according to your comments. That might be a good idea, but I'm not entirely sure how you would like that implemented.

I don't personally see any value in combining custom configuration with the default configuration. As soon as you add your own custom configuration via `wsl.tarball.configPath` you don't need the defaults, also it is quite flexible, you can have use any method to work with Nix sources to construct the configuration.

Perhaps the current `defaultConfig` (if generated a directory instead) could be the default value of `configPath`.

K900:
It's not necessarily about custom configuration - it's about adding arbitrary files in general, like secrets or SSH keys or whatever.

terlar:
You mean generally in the tarball filesystem? Or specifically inside /etc/nixos (what this is mainly about).

I guess there is also a case of being able to configure arbitrary commands being executed in the script. Having a configuration to add steps and their actions. I guess for example if you are planning to use flakes the part of channels is unnecessary. But I think that is a bigger topic.

K900:
I mean in the tarball in general, see `make-system-tarball.nix` in nixpkgs for example.

SuperSandro2000:
@nzbr @K900 can you take a look, please?

nzbr:
> I mean in the tarball in general, see `make-system-tarball.nix` in nixpkgs for example.

I'll merge this as-is and make a new issue to add something like `extraCommands`

PR #320, opened 2023-10-29
Add DistroLauncher

nzbr:
This PR is a continuation of #94, which I had to close because GitHub does not allow changing a PR's source branch once it has been created. Please look there for all previous discussion.

Add a custom DistroLauncher. This will allow packaging NixOS-WSL as an MSIX package and upload it to the MS Store eventually

This is a custom reimplementation of DistroLauncher in C# (instead of C++, which is what the reference implementation uses)
The rewrite gives us greater flexibility in extending the DistroLauncher with additional features, e.g. upgrading to newer releases of NixOS-WSL with something like `nixos update`

**Postponed** (things that will be done in separate PRs later)
- Embed version info in channel (in a separate PR)
- Re-do the update process
- Rescue Shell
- Rollback

Related: #63 #78


rogusdev:
Is this still being worked on, or has it been dropped in favor of something else? If it's still underway, is there any particular way to help / instructions url from MS on what needs to happen for remaining steps or anything? Thanks :)

nzbr:
It's still underway. It was on hold because we moved to loading the wsl-module from a channel instead of `/etc/nixos/nixos-wsl`, meaning that I'd have to reimplement the updater included in the launcher. I should probably just rip it out for now and reimplement it later, when I have more time for it, so that we can at least have the launcher merged. Once the launcher is merged, it, from what I understand, in theory be possible to push the resulting `.msixbundle` onto the MS store

Launcher/Launcher/src/Helpers/VersionHelper.cs
```diff
@@ -0,0 +1,60 @@
+using System.Reflection;
+using System.Text.Json;
+using WSL;
+
+namespace Launcher.Helpers;
+
+public static class VersionHelper {
+    public static Version? LauncherVersion => Assembly.GetEntryAssembly()?.GetName().Version;
```
thexyno, line 8:
```suggestion
    public static Version? LauncherVersion => type(VersionHelper).Assembly.GetName().Version;
```

see https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getexecutingassembly?view=net-8.0&redirectedfrom=MSDN#remarks

Launcher/Launcher/src/WSL/Kernel32/CloseHandle.cs
```diff
@@ -0,0 +1,12 @@
...
+using System.Security;
+
+namespace WSL.Kernel32;
+
+public static partial class Kernel32Loader {
+    [DllImport("kernel32.dll", SetLastError = true)]
+    [SuppressUnmanagedCodeSecurity]
+    [return: MarshalAs(UnmanagedType.Bool)]
+    public static extern bool CloseHandle(IntPtr hObject);
+}
```
thexyno, line 12: noeol here?

Launcher/default.nix
```diff
@@ -0,0 +1,87 @@
...
+, fetchurl ? pkgs.fetchurl
+, linkFarmFromDrvs ? pkgs.linkFarmFromDrvs
+}:
+
+let
+  fetchNuGet = { pname, version, sha256 }: fetchurl {
+    name = "${pname}-${version}.nupkg";
+    url = "https://www.nuget.org/api/v2/package/${pname}/${version}";
+    inherit sha256;
+  };
```
thexyno, line 16: It is currently recommended to use `hash` and not `sha256` for anything nixpkgs related 

https://nixos.org/manual/nixpkgs/stable/#fetchurl

thexyno (changes requested):
don't have any knowledge about windows internals, but looks good all in all

PR #324, opened 2023-11-03
Add a workaround to support VS Code's Remote WSL extension (fix #238)

### LINKED ISSUE #238: Help! After updating VSCode, I can no longer connect to NixOS though the WSL remote
You may have noticed that with the latest update to the [WSL remote extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl) (0.77.0), connection to NixOS fails.
This is because Micrsoft removed the option for setting the remote connection method. Where before you could get it to connect by setting the option to something other than `wslExeProxy` that is the only available option now.

There are two issues tracking possible solutions:
  - microsoft/vscode-remote-release#8305 asking them to patch the alternate connection methods back in
  - #222 here for patching NixOS on WSL to support `wslExeProxy`

Atry:
This PR adds `modules/vscode-remote-wsl-extension-workaround.nix`, when `wsl.vscodeRemoteWslExtensionWorkaround.enable` is `true`, VS Code's Remote WSL extension can connect to NixOS, fixing #238

See https://github.com/Atry/nixos-wsl-vscode/blob/main/flake.nix for usage

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -16,6 +16,22 @@
         "type": "github"
       }
     },
+    "flake-compat_2": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1673956053,
+        "narHash": "sha256-4gtG9iQuiKITOjNQQeQIpoIB6b16fm+504Ch3sNKLd8=",
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "rev": "35bb57c0c8d8b62bbfd284272c928ceb64ddbde9",
+        "type": "github"
+      },
+      "original": {
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "type": "github"
+      }
+    },
     "flake-utils": {
       "inputs": {
         "systems": "systems"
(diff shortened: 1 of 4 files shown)


K900:
I would prefer to see something like https://github.com/K900/vscode-remote-workaround, which simply uses nixpkgs node.js

Atry:
This PR is technically not a workaround. It just provides what vscode-server assumes (global C runtime and global coreutils).

K900:
Then neither is my approach, because it provides a functioning Node runtime.

Atry:
But it replaces vscode’s node on fly, right?

K900:
Correct.

SuperSandro2000 (changes requested):
Since this affects every program and can cause unexpected regressions I am very reluctant to accept it, especially regarding that we would still need an older WSL release which is probably a separate problem but is helping the situation not.

Atry:
Since you are reluctant to accept this PR. I just put it in a separate repository and close this PR.

Feel free to use this module from https://github.com/Atry/nixos-wsl-vscode

Issue #338, opened 2023-11-15
The new tarballBuilder fails when HOME variable set on execution host


terlar:
## Bug description
I discovered on my CI that sets a HOME variable that the `tarballBuilder` fails with:
```
error: opening '/github/home/.nix-channels' for writing: No such file or directory
Error: Process completed with exit code 1.
```

This is because my GitHub Runner has set this HOME variable.

## To Reproduce
```
sudo HOME=/home/unknown nix run --print-build-logs github:nix-community/NixOS-WSL#nixosConfigurations.modern.config.system.build.tarballBuilder
```

**Expected behavior**
Successful build of tarballBuilder.

## Logs
```
Include relevant console logs
```

error: opening '/github/home/.nix-channels' for writing: No such file or directory
Error: Process completed with exit code 1.

terlar:
This could be fixed like this:
```patch
diff --git a/modules/build-tarball.nix b/modules/build-tarball.nix
index 3d787ab..a53ac1a 100644
--- a/modules/build-tarball.nix
+++ b/modules/build-tarball.nix
@@ -76,7 +76,7 @@ in
           --substituters ""

         echo "[NixOS-WSL] Adding channel..."
-        nixos-enter --root "$root" --command 'nix-channel --add https://github.com/nix-community/NixOS-WSL/archive/refs/heads/main.tar.gz nixos-wsl'
+        nixos-enter --root "$root" --command 'HOME=/root nix-channel --add https://github.com/nix-community/NixOS-WSL/archive/refs/heads/main.tar.gz nixos-wsl'

         echo "[NixOS-WSL] Adding default config..."
         ${if cfg.configPath == null then ''
```

Issue #337, opened 2023-11-14
The new tarballBuilder trap fails when trying to remove `/var/empty`.


terlar:
## Bug description
I recently tried migrating over to the new `tarballBuilder`. However when I try to execute the script I get the following error:
```
rm: cannot remove '/tmp/nixos-wsl-tarball.H3cpgOr2Np/var/empty': Operation not permitted
```

Apparently that folder is marked as immutable, even preventing root from removing it, however it can be solved by:
```
chattr -i /tmp/nixos-wsl-tarball.H3cpgOr2Np/var/empty
```

## To Reproduce

```
sudo nix run --print-build-logs github:nix-community/NixOS-WSL#nixosConfigurations.modern.config.system.build.tarballBuilder
```

**Expected behavior**
Not failing on the trap after script execution.

## Logs
```
Include relevant console logs
```
...
installing the boot loader...
setting up /etc...
setting up /bin...
setting up /bin/login...
setting up /sbin/init shim...
installation finished!
[NixOS-WSL] Adding channel...
setting up /etc...
setting up /bin...
setting up /bin/login...
setting up /sbin/init shim...
[NixOS-WSL] Adding default config...
[NixOS-WSL] Compressing...
rm: cannot remove '/tmp/nixos-wsl-tarball.xXcwgyysE7/var/empty': Operation not permitted

terlar:
I got it working like this:
```patch
diff --git a/modules/build-tarball.nix b/modules/build-tarball.nix
index 8341f7f..4cdd3e5 100644
--- a/modules/build-tarball.nix
+++ b/modules/build-tarball.nix
@@ -64,7 +64,7 @@ in

         root=$(mktemp -p "''${TMPDIR:-/tmp}" -d nixos-wsl-tarball.XXXXXXXXXX)
         # FIXME: fails in CI for some reason, but we don't really care because it's CI
-        trap 'rm -rf "$root" || true' INT TERM EXIT
+        trap 'chattr -Rf -i "$root" || true && rm -rf "$root" || true' INT TERM EXIT

         chmod o+rx "$root" "''${TMPDIR:-/tmp}" -d nixos-wsl-tarball.XXXXXXXXXX)
         # FIXME: fails in CI for some reason, but we don't really care because it's CI
-        trap 'rm -rf "$root" || true' INT TERM EXIT
+        trap 'chattr -R -i "$root" && rm -rf "$root" || true' INT TERM EXIT

         chmod o+rx "$root"
```
[...]

PR #341, opened 2023-11-22
Fix tarball builder

terlar:
This fixes the tarball builder when trying to remove `/var/empty` and environments where `HOME` variable is set.

See #337 and #338


modules/build-tarball.nix
```diff
@@ -56,7 +56,7 @@ in
 
         root=$(mktemp -p "''${TMPDIR:-/tmp}" -d nixos-wsl-tarball.XXXXXXXXXX)
         # FIXME: fails in CI for some reason, but we don't really care because it's CI
-        trap 'rm -rf "$root" || true' INT TERM EXIT
+        trap 'chattr -Rf -i "$root" || true && rm -rf "$root" || true' INT TERM EXIT
```
SuperSandro2000, line 59: I don't think this is doing what you thought it would do

```suggestion
        trap 'chattr -Rf -i "$root"; rm -rf "$root"; true' INT TERM EXIT
```
we probably want something like the above

PR #364, opened 2023-12-15
Disable bootspec

SuperSandro2000:
which holds a reference to the kernel

On my system this results in the following change:

```
System package diff:
linux: 6.6.7 → ∅, -128630.1 KiB
```


K900 (approved):
LGTM

Issue #362, opened 2023-12-15
WSLg desktop integration does not show apps in Start menu and correct icons for GUI apps


mprasil:
## Problem
No applications appear in Windows Start menu (like they do for Ubuntu for example) and when I launch GUI apps in NixOS WSL, the windows get generic "tux" icons in the windows taskbar.

The problem is that the WSLg layer looks up the *.desktop files in these folders:
- /usr/share/applications
- /usr/local/share/applications
- /var/lib/snapd/desktop/applications
- /var/lib/flatpak/exports/share/applications

None of which exists in NixOS by default. 

## Solution

I have created a symlink to per-user profile for the default WSL user with something like this:

```nix
systemd.tmpfiles.settings.wsl = {                                           
  "/usr/share/applications".L = {                                           
    argument = "/etc/profiles/per-user/${wsl_user.name}/share/applications";
  };                                                                        
  "/usr/share/icons".L = {                                                  
    argument = "/etc/profiles/per-user/${wsl_user.name}/share/icons";       
  };                                                                        
};                                                                          
```

In the above code, I'm also linking the icons folder for WSLg to pick up icons correctly. This makes WSLg pick up the desktop files user might have installed:

```
# relevant parts of /mnt/wslg/weston.log
...
[09:24:57.906] app_list_monitor_thread: inotify_add_watch(/usr/share/applications)
[09:24:57.906] app_list_monitor_thread: monitor /usr/share/applications
...
[09:24:57.926] desktop file: /usr/share/applications/emacs.desktop
[09:24:57.926]     Name[en_US]:Emacs (NixOS)
[09:24:57.926]     Exec:emacs
[09:24:57.926]     TryExec:(null)
[09:24:57.926]     WorkingDir:(null)
[09:24:57.926]     Icon name:emacs
[09:24:57.926]     Icon SVG :0
[09:24:57.926]     Icon file:/usr/share/icons/hicolor/128x128/apps/emacs.png
[09:24:57.926]     Icon image:0x7f39c800bdb0
[09:24:57.926] app list entry updated: Key:emacs, Name:Emacs (NixOS)
...
```

This makes the application appear in Windows Start menu:

![image](https://github.com/nix-community/NixOS-WSL/assets/8255364/a3c46704-ad23-4ff9-9434-43007a470a63)

And also makes the windows show up with correct icon in task bar:

[...]

PR #365, opened 2023-12-17
Improve WSLg desktop integration

mprasil:
The problem is that the WSLg layer looks up the *.desktop files in these folders:

* /usr/share/applications
* /usr/local/share/applications
* /var/lib/snapd/desktop/applications
* /var/lib/flatpak/exports/share/applications

None of which exists in NixOS by default.

This adds a symlink to per-user profile for the default WSL user. With the symlinks in place WSLg can find *.desktop files and show up GUI apps in Windows Start menu.

Symlinking icons also makes sure correct icons appear for the Start menu entries and running apps.

This resolves https://github.com/nix-community/NixOS-WSL/issues/362

Note that AFAIK the WSLg applications folder discovery happens just after start of the WSL distribution, so this might require WSL restart after this has been applied for WSLg to pick up the folders/apps.

### DIFF
--- a/modules/systemd/default.nix
+++ b/modules/systemd/default.nix
@@ -41,6 +41,14 @@ with lib; {
         # Link the X11 socket into place. This is a no-op on a normal setup,
         # but helps if /tmp is a tmpfs or mounted from some other location.
         tmpfiles.rules = [ "L /tmp/.X11-unix - - - - ${cfg.wslConf.automount.root}/wslg/.X11-unix" ];
+
+        # Link *.desktop files and icons where WSLg expects them so that GUI apps show up in Start menu
+        tmpfiles.settings.wslg_integration = {
+          "/usr/share/applications".L.argument =
+            "/etc/profiles/per-user/${cfg.defaultUser}/share/applications";
+          "/usr/share/icons".L.argument =
+            "/etc/profiles/per-user/${cfg.defaultUser}/share/icons";
+        };
       };
 
     };
--- a/modules/systemd/default.nix
+++ b/modules/systemd/default.nix
@@ -44,10 +44,8 @@ with lib; {
 
         # Link *.desktop files and icons where WSLg expects them so that GUI apps show up in Start menu
         tmpfiles.settings.wslg_integration = {
-          "/usr/share/applications".L.argument =
-            "/etc/profiles/per-user/${cfg.defaultUser}/share/applications";
-          "/usr/share/icons".L.argument =
-            "/etc/profiles/per-user/${cfg.defaultUser}/share/icons";
+          "/usr/share/applications".L.argument = "/etc/profiles/per-user/${cfg.defaultUser}/share/applications";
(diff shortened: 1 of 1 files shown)


modules/systemd/default.nix
```diff
@@ -41,6 +41,14 @@ with lib; {
         # Link the X11 socket into place. This is a no-op on a normal setup,
         # but helps if /tmp is a tmpfs or mounted from some other location.
         tmpfiles.rules = [ "L /tmp/.X11-unix - - - - ${cfg.wslConf.automount.root}/wslg/.X11-unix" ];
+
+        # Link *.desktop files and icons where WSLg expects them so that GUI apps show up in Start menu
+        tmpfiles.settings.wslg_integration = {
+          "/usr/share/applications".L.argument =
+            "/etc/profiles/per-user/${cfg.defaultUser}/share/applications";
+          "/usr/share/icons".L.argument =
+            "/etc/profiles/per-user/${cfg.defaultUser}/share/icons";
```
SuperSandro2000, line 50:
```suggestion
          "/usr/share/applications".L.argument = "/etc/profiles/per-user/${cfg.defaultUser}/share/applications";
          "/usr/share/icons".L.argument = "/etc/profiles/per-user/${cfg.defaultUser}/share/icons";
```

SuperSandro2000 (commented):
didn't try it out but sounds logical to me

modules/systemd/default.nix
mprasil: Thanks for suggestion @SuperSandro2000, that's applied now.

Out of curiosity is that some specific editor config? My current setup gave me the formatting from my initial commit and I always found it a bit too eager to fold the longer lines, but I'm new enough with Nix. 😅 My current setup is essentially just default Nix config in Doom Emacs, which isn't ideal, but I haven't seen enough Nix code out there to really question it.

SuperSandro2000 (approved):

modules/systemd/default.nix
SuperSandro2000: That uses nixfmt which by default has the dated ideas about line length. My monitor is wide enough that I don't need to desperately squeeze 80 characters into one line and cribble readability.

SuperSandro2000:
Isn't this duplicated with https://github.com/nix-community/NixOS-WSL/blob/34eda458bd3f6bad856a99860184d775bc1dd588/modules/wsl-distro.nix#L106C1-L122 ?

modules/systemd/default.nix
mprasil: Yeah keeping the line short by sacrificing readability is not great. Do you use some other formatter?

mprasil:
@SuperSandro2000 you're right, that implementation actually solves this problem even though it's still broken for me. But I think I know why and will submit separate PR to fix it.

modules/systemd/default.nix
SuperSandro2000: Usually nixpkgs-fmt but adhoc changes done to existing files I usually do by hand to reduce the noise.

PR #368, opened 2023-12-18
Fix symlink creation of per-user apps/icons folder

### LINKED ISSUE #362: WSLg desktop integration does not show apps in Start menu and correct icons for GUI apps

mprasil:
Closes #362

The code relied on `defaultUser` attribute name in `config.users.users` to be the same as the username. This is not always the case as the username can be set by `name` attribute inside the user configuration

So for following user

```nix
users.users.foo.name = "bar";
```

the `wsl.defaultUser = "foo"` works, but the `wsl.startMenuLaunchers = true` won't work because it tries to find applications folder at `/etc/profiles/per-user/foo/share` rather than
`/etc/profiles/per-user/bar/share` that is actually created.

This PR uses the `name` attribute of that user instead which should fix the issue. (tested locally, it seems to be working correctly)

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -109,8 +109,8 @@ in
             if [[ -d "$systemConfig/sw/share/$x" ]]; then
               targets+=("$systemConfig/sw/share/$x/.")
             fi
-            if [[ -d "/etc/profiles/per-user/${cfg.defaultUser}/share/$x" ]]; then
-              targets+=("/etc/profiles/per-user/${cfg.defaultUser}/share/$x/.")
+            if [[ -d "/etc/profiles/per-user/${config.users.users.${cfg.defaultUser}.name}/share/$x" ]]; then
+              targets+=("/etc/profiles/per-user/${config.users.users.${cfg.defaultUser}.name}/share/$x/.")
             fi
 
             if (( ''${#targets[@]} != 0 )); then


SuperSandro2000 (approved):

terlar:
This looks good. I just tried to figure out why my icons were not working and just found the solution:
```
diff --git a/modules/wsl-distro.nix b/modules/wsl-distro.nix
index 50b81ea..39aae15 100644
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -125,7 +125,7 @@ in

             if (( ''${#targets[@]} != 0 )); then
               mkdir -p "/usr/share/$x"
-              ${pkgs.rsync}/bin/rsync -ar --delete-after "''${targets[@]}" "/usr/share/$x"
+              ${pkgs.rsync}/bin/rsync -ar --copy-dirlinks --delete-after "''${targets[@]}" "/usr/share/$x"
             else
               rm -rf "/usr/share/$x"
             fi
```

Since the `hicolor` icons folder was a symlink in the derivation rsync couldn't merge the sources and hence just copied one of them. This change ensures it copies directory links instead to avoid such issues. Tested and working with icons for me now. I'll wait with PR until this is merged.

SuperSandro2000:
I've created #381 with your suggestion

Issue #266, opened 2023-07-12
Make WSLg support optional


nzbr:
Everything related to GUI-Support should be separated out into its own module. 
While the module should be enabled by default, it should be possible to disable it in order to reduce the system's footprint for people who don't want/need it

kjhoerr:
Yes please! I can't use WSLg for reasons outside of my control and all the gui support derivations add a lot of time to what should be a pretty flat build.

flickerfly:
Agreed

inschrift-spruch-raum:
pull已经合并了,问题解决了吗

PR #371, opened 2023-12-21
feat: add option to disable wslg x11 mount

PhilippHeuer:
This PR migrates from `systemd.tmpfiles.rules` to `systemd.tmpfiles.settings` to allow user overrides.

This way the `/tmp/.X11-unix` wslg mount can be disabled with:

```nix
systemd.tmpfiles.settings = {
  "10-wslg-x11" = lib.mkForce {};
};
```

For reference the old and new content in `/etc/tmpfiles.d`:

old: `L /tmp/.X11-unix - - - - /mnt/wslg/.X11-unix`
new: `'L' '/tmp/.X11-unix' '-' '-' '-' '-' /mnt/wslg/.X11-unix`

### DIFF
--- a/modules/systemd/default.nix
+++ b/modules/systemd/default.nix
@@ -12,6 +12,11 @@ with lib; {
       default = true;
       description = "Use native WSL systemd support";
     };
+    wslgMountX11 = mkOption {
+      type = bool;
+      default = true;
+      description = "Mount the WSLg .X11-unix directory into /tmp";
+    };
   };
 
   config =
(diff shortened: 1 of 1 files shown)


modules/systemd/default.nix
```diff
@@ -40,7 +45,7 @@ with lib; {
 
         # Link the X11 socket into place. This is a no-op on a normal setup,
         # but helps if /tmp is a tmpfs or mounted from some other location.
-        tmpfiles.rules = [ "L /tmp/.X11-unix - - - - ${cfg.wslConf.automount.root}/wslg/.X11-unix" ];
+        tmpfiles.rules = mkIf cfg.wslgMountX11 ([ "L /tmp/.X11-unix - - - - ${cfg.wslConf.automount.root}/wslg/.X11-unix" ]);
```
SuperSandro2000, line 48: The () here are not required if I eyeballed that correct.

Also can't we use the new tmpfiles syntax to easily disable this with mkForce? I would prefer that over a new setting.

nzbr: Agreed, using the new tmpfiles syntax from now on sounds like a good Idea. 
There should be a single setting to disable everything related to WSLg (see #266). If there's a need to disable the tmpfiles rule on its own, it could still be overriden with mkForce as @SuperSandro2000 said

PhilippHeuer: Thanks for the Feedback, i didn't know about the new tmpfiles syntax.

I updated the PR and tested it, overriding works fine so this is a better solution.

nzbr:
Thanks @PhilippHeuer!

PR #380, opened 2024-01-03
fix: start menu launchers rsync of directory symlinks

terlar:
When using a profile containing icons having a symlink such as:
```
/etc/profiles/per-user/USER/share/icons/hicolor -> /nix/store/fvlvfdkc9hh7icadwqxfiazifmsxprrl-home-manager-path/share/icons/hicolor
```

The merge of several rsync sources doesn't work properly. With this fix it will copy the symlinked directory instead and in that way support the rsync.

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -125,7 +125,7 @@ in
 
             if (( ''${#targets[@]} != 0 )); then
               mkdir -p "/usr/share/$x"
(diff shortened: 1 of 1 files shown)


terlar:
This doesn't have conflicts and didn't touch the same lines as the other change after all, so I guess no need to rebase.

SuperSandro2000 (approved):

PR #339, opened 2023-11-16
wsl-distro: introduce `wsl.useWindowsDriver` option

Atry:
This PR introduce `wsl.useWindowsDriver` option, which enables `/usr/lib/wsl/lib` as OpenGL driver.

With the help of this PR, CUDA can be used if there is a Nvidia GPU.

```
[nixos@nixos:~]$ nvidia-smi
Thu Nov 16 05:46:07 2023       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 545.23.05              Driver Version: 545.84       CUDA Version: 12.3     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA GeForce RTX 3060        On  | 00000000:07:00.0  On |                  N/A |
|  0%   38C    P8              18W / 170W |   1481MiB / 12288MiB |      2%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
                                                                                         
+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|    0   N/A  N/A        25      G   /Xwayland                                 N/A      |
+---------------------------------------------------------------------------------------+
```

PR #382, opened 2024-01-04
Fix useWindowsDriver when other opengl driver are used like intel-compute-runtime

SuperSandro2000:
see https://github.com/nix-community/NixOS-WSL/pull/339#discussion_r1441907386

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -76,8 +76,11 @@ in
 
       extraPackages = mkIf cfg.useWindowsDriver [
         (pkgs.runCommand "wsl-lib" { } ''
-          mkdir "$out"
-          ln -s /usr/lib/wsl/lib "$out/lib"
+          mkdir -p "$out/lib"
+          # we cannot just symlink the lib directory because it breaks merging with other drivers that provide the same directory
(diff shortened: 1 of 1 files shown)


modules/wsl-distro.nix
```diff
@@ -76,8 +76,11 @@ in
 
       extraPackages = mkIf cfg.useWindowsDriver [
         (pkgs.runCommand "wsl-lib" { } ''
-          mkdir "$out"
-          ln -s /usr/lib/wsl/lib "$out/lib"
+          mkdir -p "$out/lib"
+          # we cannot just symlink the lib directory because it breaks merging with other drivers that provide the same directory
+          for so in libcuda.so libd3d12.so libd3d12core.so libdxcore.so; do
```
SuperSandro2000, line 81: Do we need other cuda so's? What are the amd names?

Atry: AMD does not support CUDA

SuperSandro2000: I meant other shared objects in general, like the amd one. What's the name for it? Is there even one?

SuperSandro2000:
see #383

PR #383, opened 2024-01-04
Hard-code file names when making symbolic links for drivers

Atry:
This PR addresses https://github.com/nix-community/NixOS-WSL/pull/339#discussion_r1441907386

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -77,7 +77,24 @@ in
       extraPackages = mkIf cfg.useWindowsDriver [
         (pkgs.runCommand "wsl-lib" { } ''
           mkdir "$out"
-          ln -s /usr/lib/wsl/lib "$out/lib"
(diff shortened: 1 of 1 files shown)


modules/wsl-distro.nix
```diff
@@ -77,7 +77,24 @@ in
       extraPackages = mkIf cfg.useWindowsDriver [
         (pkgs.runCommand "wsl-lib" { } ''
           mkdir "$out"
-          ln -s /usr/lib/wsl/lib "$out/lib"
+          ln -s /usr/lib/wsl/lib/libcudadebugger.so.1 "$out/lib"
```
SuperSandro2000, line 80:
```suggestion
          # # we cannot just symlink the lib directory because it breaks merging with other drivers that provide the same directory
          ln -s /usr/lib/wsl/lib/libcudadebugger.so.1 "$out/lib"
```

SuperSandro2000 (approved):

PR #398, opened 2024-01-11
make /etc/nixos writable

### LINKED ISSUE #384: Make configuration.nix writeable by root
When importing a fresh tarball, `/etc/nixos/configuration.nix` is set to read-only. We should make the file writeable by root in the tarball builder

aikooo7:
# What is the objective of this PR

- make /etc/nixos writable

# Note

I based myself in the welcome script and tested it before committing

Closes #384

### DIFF
--- a/modules/build-tarball.nix
+++ b/modules/build-tarball.nix
@@ -23,6 +23,19 @@ let
       wsl.defaultUser = "nixos";
       ${lib.optionalString (!config.wsl.nativeSystemd) "wsl.nativeSystemd = false;"}
 
+      system.activationScripts.NixOS-WSL-Permissions-Fixer.text =
+      '''
+        if [ -f /etc/nixos/.permissions-fixer-needed ]; then
+          echo "[NixOS-WSL-Permissions-Fixer] Making /etc/nixos writable to current user..."
+          ${pkgs.acl}/bin/setfacl -R -m u:nixos:rwx /etc/nixos
+          if [ "$?" -eq 0 ]; then
+            rm /etc/nixos/.permissions-fixer-needed
+          fi
+        else
+          echo "[NixOS-WSL-Permissions-Fixer] Permissions don't need to be fixed."
+        fi
+      ''';
+
       # This value determines the NixOS release from which the default
       # settings for stateful data, like file locations and database versions
       # on your system were taken. It's perfectly fine and recommended to leave
@@ -86,6 +99,7 @@ in
           cp -R ${lib.cleanSource cfg.configPath}/. "$root/etc/nixos"
           chmod -R u+w "$root/etc/nixos"
         ''}
+        touch "$root/etc/nixos/.permissions-fixer-needed"
 
         echo "[NixOS-WSL] Compressing..."
         tar -C "$root" \


aikooo7:
I had my email wrong so that is the reason the commit was unverified and that is also why I did a force-push

aikooo7:
Just another note: if wsl2 is runned too much fast and you get a error that bash shouldn't return and you run the wsl2 again you will get a error saying that sudo uid needs to 0, so you need to close and open wsl2 again to it work.
My assumption is that is because of the native systems because in my daily driver when I turned off it, I stopped getting the return error.

**Sorry for the bad English, just didn't want this pr to be merged while I am not sure why this happens, when I go back into my setup I while provide all the errors and re write this comment**

SuperSandro2000:
> I had my email wrong so that is the reason the commit was unverified and that is also why I did a force-push

I don't think anyone cares here. Feel free to force push the review comments, too.

aikooo7:
Testing again my implementation I found out that I almost every time in boot I get the error: ``sudo: /run/current-system/sw/bin/sudo must be owned by uid 0 and have the setuid bit set``, but the thing is I am able to run root and nixos-wsl-permissions-fix after the shell loaded, so my assumption is that somehow in the first boot the sudo permissions are not correct at the time I run the script.

I'm really out of ideas in how I could make it work

nzbr:
Instead of using shell init, you could let this run as an activation script. That way, you don't need sudo. Otherwise, you need to use sudo from `/run/wrappers/bin`. Files in the nix store can not have the suid bit set, therefore NixOS creates wrappers for them through an activation script

nzbr:
Only saw your comment on the issue now, sorry. The activation script could be handled the same way the shell hook currently is - so that it's gone after the first rebuild. Another option would be to create a file in the tarball builder, that gets removed by the script, so that the chmod only runs once

aikooo7:
> activation script

Using /run/wrappers/bin/sudo gives me: ``/run/current-system/sw/bin/nixos-wsl-permissions-fix: line 3: /run/wrappers/bin/sudo: No such file or directory``

Using a activation script will always change permissions, and that may not be what the users want, but if you want me to do it I can for sure

aikooo7:
> Only saw your comment on the issue now, sorry. 

Dont' worry about that, it happens with everyone

> Another option would be to create a file in the tarball builder, that gets removed by the script, so that the chmod only runs once

I tried doing ``nixos-enter --root "$root" --command 'se^Cacl -R -m u:nixos:rwx /etc/nixos'`` and I got ``chmod: changing permissions of '/dev': Operation not permitted
chmod: changing permissions of '/sys': Operation not permitted``

The thing is, even if doing a command like pwd I get the same think

aikooo7:
> > activation script
> 
> Using /run/wrappers/bin/sudo gives me: `/run/current-system/sw/bin/nixos-wsl-permissions-fix: line 3: /run/wrappers/bin/sudo: No such file or directory`
> 
> Using a activation script will always change permissions, and that may not be what the users want, but if you want me to do it I can for sure

Do you want me to use a activation script? @nzbr

aikooo7:
> Either that or a systemd service. Having it as a part of the shell init is suboptimal imho. Make the tarball builder touch a file like `/etc/nixos/.chmod` or something and check if it exists. If it does, `rm` the file and run the `chmod`. That way we can even put it in the regular config so that it still works for people who build their own tarballs

I also had idea of the builder touching a file but I thought it was bad practice. Since it is okay I will implement it.

aikooo7:
In my machine it works perfectly, but I wonder, giving executable permissions to it is fine or it may be a security risk?

modules/build-tarball.nix
```diff
@@ -23,6 +23,19 @@ let
       wsl.defaultUser = "nixos";
       ${lib.optionalString (!config.wsl.nativeSystemd) "wsl.nativeSystemd = false;"}
 
+      system.activationScripts.NixOS-WSL-Permissions-Fixer.text =
+      '''
+        if [ -f /etc/nixos/.permissions-fixer-needed ]; then
+          echo "[NixOS-WSL-Permissions-Fixer] Making /etc/nixos writable to current user..."
+          ${pkgs.acl}/bin/setfacl -R -m u:nixos:rwx /etc/nixos
```
aikooo7, line 30: Giving executable permissions to it is fine or it may be a security risk?

nzbr: Explicitly setting the executable bit shouldn't be necessary. The directory should have it set already (needed to ls it) and the files in there don't need it.
Two other things: Please move the whole thing to `wsl-distro.nix` instead of the config template. Besides that i'd rather have the files writeable by root instead of the default user (who might have a user name other than nixos anyway). With that using the `acl` package shouldn't be necessary. `${pkgs.coreutils}/chmod -R u+w` would be sufficient.
Rest looks good to me

aikooo7: > Besides that i'd rather have the files writeable by root instead of the default user (who might have a user name other than nixos anyway). With that using the `acl` package shouldn't be necessary. `${pkgs.coreutils}/chmod -R u+w` would be sufficient. Rest looks good to me

Files is already writable by root (tested in lastest release), I wonder what we are trying to fix then?

nzbr: You're right, I just double-checked. I was convinced it was still `-r--r--r--`, but apparently that _was_ fixed already. I am so sorry for having wasted your time...

aikooo7: Don't worry about that, it helped me learn some very cool stuff!

aikooo7:
I will be closing this since it was already fixed

PR #406, opened 2024-01-25
More documentation

### LINKED ISSUE #250: Document debugging tips
## Problem
Using Windows when I'm a NixOS native, I was having permission issues. I fixed the problem by updating WSL.

## Solution
`wsl --update`

I'm opening this issue to suggest documenting common pitfalls like this. Either as a section in the README or a seperate markdown file. At the very least this issue can now be found when troubleshooters search for 'permission'.

:clinking_glasses:

nzbr:
Closes #250

### DIFF
--- a/docs/src/how-to/change-username.md
+++ b/docs/src/how-to/change-username.md
@@ -1,0 +1,19 @@
+# How to change the username
+
+If you want to change the default username to something other than `nixos`, use the `wsl.defaultUser` option.
+When building your own tarball, this should be sufficient. A user with the name specified in that option will be created automatically.
+
+Changing the username on an already installed system is possible as well.
+Follow these instructions to make sure, the change gets applied correctly:
+
+1. Change the `wsl.defaultUser` setting in your configuration to the desired username.
+2. Apply the configuration:\
+   `nixos-rebuild boot`\
+   Do not use `nixos-rebuild switch`! It may lead to the new user account being misconfigured.
+3. Exit the WSL shell and stop your NixOS distro:\
+   `wsl -t NixOS`.
+4. Start a shell inside NixOS and immediately exit it to apply the new generation:\
+   `wsl -d NixOS --user root exit`
+5. Stop the distro again:\
+   `wsl -t NixOS`
+6. Open a WSL shell. Your new username should be applied now!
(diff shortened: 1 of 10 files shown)


docs/src/how-to/change-username.md
```diff
@@ -0,0 +1,19 @@
...
+If you want to change the default username to something other than `nixos`, use the `wsl.defaultUser` option.
+When building your own tarball, this should be sufficient. A user with the name specified in that option will be created automatically.
+
+Changing the username on an already installed system is possible as well.
+Follow these instructions to make sure, the change gets applied correctly:
+
+1. Change the `wsl.defaultUser` setting in your configuration to the desired username.
+2. Apply the configuration:\
+   `nixos-rebuild boot`\
+   Do not use `nixos-rebuild switch`! It may lead to the new user account being misconfigured.
```
SuperSandro2000, line 12:
```suggestion
2. Apply the configuration: `nixos-rebuild boot` Do not use `nixos-rebuild switch`! It may lead to the new user account being misconfigured.
```
Is this equal to this?

nzbr: No, the backslash makes it so that there is a newline in the rendered HTML (so exactly the opposite of nearly everywhere else)

docs/src/how-to/change-username.md
```diff
@@ -0,0 +1,19 @@
...
+
+If you want to change the default username to something other than `nixos`, use the `wsl.defaultUser` option.
+When building your own tarball, this should be sufficient. A user with the name specified in that option will be created automatically.
+
+Changing the username on an already installed system is possible as well.
+Follow these instructions to make sure, the change gets applied correctly:
+
+1. Change the `wsl.defaultUser` setting in your configuration to the desired username.
+2. Apply the configuration:\
+   `nixos-rebuild boot`\
```
dersuchmann, line 11:
```suggestion
   `sudo nixos-rebuild boot`\
```
I suggest either writing it like this or adding a note that you need to be root / sudo'ing for this. I edited my nixos config via `sudo nano /etc/nixos/configuration.nix` and then I got an error applying this command.

nzbr: Done, thanks for the input!

nzbr:
Thanks, I extended the --shutdown point in the troubleshooting guide a bit further, adding that it will stop other distros as well and that `wsl -t <distro>` might be a first replacement in case that is not desirable at the moment

PR #438, opened 2024-03-11
Update README.md - use %localappdata% instead of the cwd

noamraph:
Hi,

I first installed NixOS-WSL according to the instructions in the README.md, and it created a NixOS folder in my Downloads folder, in which the filesystem was stored. I saw that the other WSL filesystems are located in %localappdata% (AppData\Local), so this seems like a much better instruction to give in the README file.

Also, the README include adding `--version 2`. It works, but I don't think it does anything. `wsl --help` shows that `--version` just displays the version. I checked, and the command works without the `--version 2` arguments just as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Modules for running NixOS on the Windows Subsystem for Linux
 2. Import the tarball into WSL:
 
 - ```powershell
-  wsl --import NixOS %localappdata%\NixOS\ nixos-wsl.tar.gz
+  wsl --import NixOS %userprofile%\NixOS\ nixos-wsl.tar.gz
   ```
 
 3. You can now run NixOS:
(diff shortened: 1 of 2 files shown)


SuperSandro2000:
> I saw that the other WSL filesystems are located in %localappdata% (AppData\Local), so this seems like a much better instruction to give in the README file.

That's because they are installed through the store and managed by the respective apps. Since we want to add that eventually, too, we should not occupy that directory. So far I always placed it under my user directory.

noamraph:
@SuperSandro2000 I agree, the user directory seems like a better place. So, I changed the PR to use `%userprofile%` instead of `%localappdata%`.

nzbr:
For consistency, please change that in `docs/src/install.md` as well

noamraph:
@nzbr thanks, I didn't know about that file. Done!

nzbr:
FYI: I changed the primary install command to PowerShell, because that's the default shell on Windows now. The detailed instructions now have both commands

noamraph:
Great, thanks!

Issue #441, opened 2024-03-20
How-To Doc


JasonGoemaat:
Very new to Nix, love this repository but had a hard time figuring out how to just add packages to the global `configuration.nix`.  I saw there was a space for a howto doc that wasn't created so I decided to add some examples.

0xLunch:
You'll want to do a bunch of googling on different setups since there are multiple ways to get your programs installed. I'll give a quick rundown here but this is more a question for Nix communities directly and not specific to NixOS-WSL.

For global system-wide packages that will exist for every user, every system, etc you can use `environment.systemPackages` inside your `etc/nixos/configuration.nix` file
example:
```nix
  # System Packages
  environment.systemPackages = with pkgs; [
    zsh
    wget
    htop
    tmux
  ];
```

When you `sudo nixos-rebuild switch` these will be included systemwide.

Another way is to simply enable a program that's in nixpkgs, ie: `programs.zsh.enable = true`. This can be done in your general configuration, or more preferably within your `home-manager` config so it applies to your specific user and not the entire system.

Hope this points you in the right direction.

nzbr:
This is basically a duplicate of the PR, so I'm closing this issue

PR #442, opened 2024-03-20
Added how to doc showing nix-shell and editing config

JasonGoemaat:
#441

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -7,7 +7,7 @@
 - [Installation](./install.md)
 - [Design](./design.md)
 - [Building](./building.md)
-- [How-To]()
+- [How-To](./how-to.md)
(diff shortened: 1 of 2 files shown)


ruebenramirez (approved):

nzbr:
> Not sure if this is what should have been placed here. Those are all normal NixOS things and we can just link to some upstream doc about this.

Agreed, I think we should keep our docs to WSL specific things and point to some general NixOS guide

aikooo7:
My 2 cents here but I also agree with the others. If we allowed this then we would need to keep it update with another documentation, why not just let the user open it by himself? Since they are using wsl2 it is assumed they can open the documentation with internet or pre-download it before.

nzbr:
I'm closing this for now. Feel free to open a new PR for linking to the correct place in the NixOS docs from ours

PR #443, opened 2024-04-02
Add option to disable error about ldconfig

SuperSandro2000:


nzbr:
Is this option documented somewhere by Microsoft? I tried to look it up but couldn't find anything

SuperSandro2000:
The option was written into the error message in dmesg.

nzbr:
Given that you wrote that it doesn't work with NixOS: Should we add a warning if someone tries to use it anyway?

SuperSandro2000:
I mean, if people provide the directories and /sbin/ldconfig then it might work, not sure. I don't think we need to add a warning for this now.

PR #444, opened 2024-04-02
usbip: make more reliable by using default route

SuperSandro2000:
The nameserver does not necessarily need to be the windows host.

/cc @terlar

### DIFF
--- a/modules/usbip.nix
+++ b/modules/usbip.nix
@@ -34,11 +34,14 @@ in
         after = [ "network.target" ];
 
         scriptArgs = "%i";
-        path = [ pkgs.linuxPackages.usbip ];
+        path = with pkgs; [
+          iproute2
+          linuxPackages.usbip
+        ];
 
         script = ''
           busid="$1"
-          ip="$(grep nameserver /etc/resolv.conf | cut -d' ' -f2)"
+          ip="$(ip route list | sed -nE 's/(default)? via (.*) dev eth0 proto kernel/\2/p')"
 
           echo "Starting auto attach for busid $busid on $ip."
           source ${usbipd-win-auto-attach} "$ip" "$busid"
--- a/modules/usbip.nix
+++ b/modules/usbip.nix
@@ -1,7 +1,5 @@
 { config, lib, pkgs, ... }:
 
-with lib;
-
 let
   usbipd-win-auto-attach = pkgs.fetchurl {
     url = "https://raw.githubusercontent.com/dorssel/usbipd-win/v3.1.0/Usbipd/wsl-scripts/auto-attach.sh";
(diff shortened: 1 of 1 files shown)


terlar:
Will this work? When I use this command I get quite a bit of results, none which is the correct one.

```
$ ip route list | sed -nE 's/(default)? via (.*) dev eth0 proto kernel/\2/p' | wc -l
115
```

I am currently using `networkingMode=mirrored`, which is not working with the current implementation either. Then I need to use `127.0.0.1` or if I use `wsl-vpnkit` I have to use `192.0.0.1`.

SuperSandro2000:
> Then I need to use `127.0.0.1`

We can add a condition for that. I don't think I want to add anything related to wsl-vpnkit as it got redundant with the network changes recently.

> Would be nice to have a solution that worked with all the various setups. Perhaps make it a configuration option if there is no reliable way to get the correct IP.

There is none and the IP isn't hardcoded by default, hence we cannot hardcode it.

SuperSandro2000:
I've added an option to allow to customize the snippet to obtain the ip address. It is also possible to just hardcode an IP. This should suffice for all listed usecases even if it involves a little bit of manual work.

terlar:
That works for me, currently I used this one:
```
busid="$1"

wsl2_gateway_ip=192.0.0.1
ip="127.0.0.1"

if ping -q -w 1 -c 1 "$wsl2_gateway_ip" >/dev/null; then
  ip="$wsl2_gateway_ip"
fi
```

I agree that we shouldn't add any custom logic to support wsl-vpnkit. Hopefully it is possible to get rid of it at some point. The VPN my company uses, still doesn't work without it, no matter what network mode/configuration is used.

I guess other option could be a list of static IP addresses and then fall-back to the lookup you have. Or just leave it totally up to the user as you have done here.

SuperSandro2000:
Since checking each IP with ping takes one second and it will increase the complexity more than I want it to right now, I will just leave it up to the end user to configure it correct.

terlar:
I tried this with NAT and it didn't work due to weird Nix string escape, not sure if it is a Nix lang regression.

```
nix-repl> "$(ip route list | sed -nE 's/(default)? via (.*) dev eth0 proto kernel/\2/p')"
"$(ip route list | sed -nE 's/(default)? via (.*) dev eth0 proto kernel/2/p')"

nix-repl> "$(ip route list | sed -nE 's/(default)? via (.*) dev eth0 proto kernel/\\2/p')"
"$(ip route list | sed -nE 's/(default)? via (.*) dev eth0 proto kernel/\\2/p')"
```

So with one escape the `\` disappears and with two `\\` it is not escaped...

terlar:
Or I guess that is just the representation of the string:
```
nix-repl> ''$(ip route list | sed -nE 's/(default)? via (.*) dev eth0 proto kernel/\2/p')''
"$(ip route list | sed -nE 's/(default)? via (.*) dev eth0 proto kernel/\\2/p')"
```

So I think it needs to be double escaped.

PR #455, opened 2024-04-17
Don't mask resolved systemd unit 

SuperSandro2000:


SuperSandro2000:
For example when using tailscale. It regularly breaks when opening new tabs and complains about changed resolve.conf. I think WSL might try to change the file under it's foot and then things don't work anymore.

PR #459, opened 2024-04-26
utils/shell-wrapper: hardcode paths to the right sh/env executables

K900:


nzbr (approved):
We could in theroy maybe save on rebuilds by not compiling the paths into the binay. But there'll probably be a rebuild  anyway when the path of `sh` changes

PR #464, opened 2024-05-05
fix(shell-wrapper): propagate login shell status

nzbr:
Starts the wrapped shell as a login shell if the shell wrapper is started as a login shell (arg0 is prefixed with `-`)

### DIFF
--- a/utils/src/shell_wrapper.rs
+++ b/utils/src/shell_wrapper.rs
@@ -47,8 +47,27 @@ fn real_main() -> anyhow::Result<()> {
         }
     }
 
+    let shell_exe = &shell
+        .file_name()
+        .ok_or(anyhow!("when trying to get the shell's filename"))?
+        .to_str()
+        .ok_or(anyhow!(
+            "when trying to convert the shell's filename to a string"
+        ))?
+        .to_string();
+
+    let arg0 = if env::args()
+        .next()
+        .map(|arg| arg.starts_with('-'))
+        .unwrap_or(false)
+    {
+        "-".to_string() + shell_exe.as_str() // prepend "-" to the shell's arg0 to make it a login shell
+    } else {
+        shell_exe.clone()
+    };
+
     Err(anyhow!(Command::new(&shell)
-        .arg0(shell)
+        .arg0(arg0)
         .args(env::args_os().skip(1))
         .exec())
     .context("when trying to exec the wrapped shell"))


SuperSandro2000:
Not working for me :/

```
May 06 16:55:15 francium systemd[1]: Stopping User Manager for UID 1000...
May 06 16:55:15 francium systemd[409]: Activating special unit Exit the Session...
May 06 16:55:15 francium ssh-agent[468]: exiting on signal 15
May 06 16:55:15 francium systemd[409]: Stopped target Main User Target.
May 06 16:55:15 francium systemd[409]: Stopping SSH Agent...
May 06 16:55:15 francium systemd[409]: Stopped SSH Agent.
May 06 16:55:15 francium systemd[409]: Stopped target Basic System.
May 06 16:55:15 francium systemd[409]: Stopped target Paths.
May 06 16:55:15 francium systemd[409]: Stopped target Sockets.
May 06 16:55:15 francium systemd[409]: Stopped target Timers.
May 06 16:55:15 francium systemd[409]: Closed D-Bus User Message Bus Socket.
May 06 16:55:15 francium systemd[409]: Closed GnuPG cryptographic agent and passphrase cache.
May 06 16:55:15 francium systemd[409]: Closed Socket for Lorri Daemon.
May 06 16:55:15 francium systemd[409]: Closed Podman API Socket.
May 06 16:55:15 francium systemd[409]: Removed slice User Application Slice.
May 06 16:55:15 francium systemd[409]: Reached target Shutdown.
May 06 16:55:15 francium systemd[409]: Finished Exit the Session.
May 06 16:55:15 francium systemd[409]: Reached target Exit the Session.
May 06 16:55:15 francium (sd-pam)[414]: pam_unix(systemd-user:session): session closed for user sandro
May 06 16:55:15 francium systemd[1]: user@1000.service: Deactivated successfully.
May 06 16:55:15 francium systemd[1]: Stopped User Manager for UID 1000.
```

nzbr:
It does not fix systemd user sessions. Only that the shell is started as a login shell.
Can you see if systemd user sessions work with another distro on your system? Because they do not for me (tested Ubuntu and Debian so far)

SuperSandro2000:
> Can you see if systemd user sessions work with another distro on your system?

They work when I go back to 0b90c1d982d443358b3f7b3a303405449a2bfe54. I don't think I enjoyed lingering or anything else. 

They also work since then but only for a few minutes and then they get closed.

nzbr:
WSL starts a single shell under `login` to create the logind session. That shell runs in the background and is not the one that gets connected to the terminal. The process has SIGCHLD set to be ignored, which makes the shell wrapper crash when it tries to wait for the `sh` it forks of to source `/etc/set-environment`.
I have changed it so that it now checks if SIGCHLD is ignored and if yes it skips setting the environment.
tl;dr: it should work now

utils/src/shell_wrapper.rs
```diff
@@ -18,37 +20,76 @@ fn real_main() -> anyhow::Result<()> {
...
+            unsafe {
+                // WSL starts a single shell under login to make sure that a logind session exists.
+                // That shell is started with SIGCHLD ignored
+                // If it is, we are probably that shell and can just skip setting the environment
+                let mut act: sigaction = MaybeUninit::zeroed().assume_init();
+                sigaction(SIGCHLD, PT_NULL as *const sigaction, &mut act);
+                if act.sa_sigaction == SIG_IGN {
+                    return Ok(());
+                }
+            }
```
K900, line 38: https://docs.rs/nix/latest/nix/sys/signal/fn.sigaction.html

utils/src/shell_wrapper.rs
```diff
@@ -58,5 +99,12 @@ fn main() {
...
 
     env::set_var("RUST_BACKTRACE", "1");
-    eprintln!("[shell-wrapper] Error: {:?}", result.unwrap_err());
+    let err = result.unwrap_err();
+
+    eprintln!("[shell-wrapper] Error: {:?}", &err);
+
+    // Write the result to /tmp/shell-wrapper_crash_<pid>.log
+    let pid = std::process::id();
+    let output_file = format!("/tmp/shell-wrapper_crash_{}.log", pid);
```
K900, line 108: Can we log to the journal instead?

utils/src/shell_wrapper.rs
K900: Never mind that can't read (by passing null), only write. Should probably add a comment for that.

nzbr: added a comment

utils/src/shell_wrapper.rs
nzbr: It was a quick and dirty implementation so I could see what was going on. Journal sounds like a way better idea. I've changed it to do that instead

PR #473, opened 2024-05-12
syschdemd: add deprecation notice

nzbr:
This should ideally be merged after #466


K900 (approved):
LGTM. Should we have a deadline? 24.11?

nzbr:
I also removed the legacy tarball from the CI pipeline now.
24.11 makes sense - it will be deprecated with 24.05 and removed one release later

K900:
Should probably add that to the message then.

nzbr:
Will do (after i fix the pipeline, whoops)

Issue #433, opened 2024-03-01
nvidia-container-toolkit-cdi-generator not work


qwqawawow:
## Bug description
nvidia-container-toolkit-cdi-generator not work.
( meanwhile Nix would install LinuxKernel and Nvidia Drivers(Linux Version))

## To Reproduce

[Nvidia CDI-support doc](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/cdi-support.html)

1.  nixpkgs 24.05 option  `virtualisation.containers.cdi.dynamic.nvidia.enable = true;`  provide [nvidia-container-toolkit-cdi-generator.service](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/hardware/nvidia-container-toolkit-cdi-generator/cdi-generate.nix#L29)

2. `systemctl status  nvidia-container-toolkit-cdi-generator.service`

## Logs
```
 /nix/store/6azwdj7ar1nlp2hn4ivfjgqd12pk0kgr-container-toolkit-container-toolkit-1.15.0-rc.3/bin/nvidia-ctk --debug cdi generate --library-search-path /usr/lib/wsl/lib/

DEBU[0000] Locating NVIDIA Container Toolkit CLI as nvidia-ctk
DEBU[0000] Locating "nvidia-ctk" in [/run/wrappers/bin /usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin /usr/games /usr/local/games /usr/lib/wsl/lib /home/nixos/.nix-profile/bin /nix/profile/bin /home/nixos/.local/state/nix/profile/bin /etc/profiles/per-user/nixos/bin /nix/var/nix/profiles/default/bin /run/current-system/sw/bin]
WARN[0000] Failed to locate nvidia-ctk: pattern nvidia-ctk not found
DEBU[0000] Using NVIDIA Container Toolkit CLI path nvidia-ctk
DEBU[0000] Is WSL-based system? false: could not load DXCore library: libdxcore.so: cannot open shared object file: No such file or directory
DEBU[0000] Is NVML-based system? false: could not load NVML library: libnvidia-ml.so.1: cannot open shared object file: No such file or directory
DEBU[0000] Is Tegra-based system? false: /sys/devices/soc0/family file not found
INFO[0000] Auto-detected mode as "nvml"
[...]

PR #478, opened 2024-05-20
fix: set wsl.useWindowsDriver when the nvidia-ctk is enabled

ereslibre:
This improves the user experience as whenever the user enables the `config.hardware.nvidia-container-toolkit.enable` option, they cannot use their Nvidia GPU's within the Docker containers because of missing libraries.

This gets fixed by setting `wsl.useWindowsDriver` explicitly when the user requests to enable GPU support on Docker containers.

Issue and fix provided by @qwqawawow

Related: https://github.com/nix-community/NixOS-WSL/issues/433
Related: https://github.com/NVIDIA/nvidia-container-toolkit/issues/452

---

Note: unsure if this is the right approach for this project. I am the maintainer of the CDI-generation for NixOS, and we got a couple of reports that users could not use their GPU's inside Docker containers.

I cannot test it since I don't use NixOS-WSL, but I thought it would be good to improve the user experience here a bit if possible.

Thanks :)

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -3,6 +3,7 @@
     ./build-tarball.nix
     ./docker-desktop.nix
     ./interop.nix
+    ./nvidia.nix
     ./recovery.nix
     ./systemd
     ./usbip.nix
--- a/modules/nvidia.nix
+++ b/modules/nvidia.nix
@@ -1,0 +1,24 @@
+{ config, lib, ... }:
+{
+
+  config =
+    lib.mkIf
+      (
+        config.wsl.enable &&
+        (config.wsl.docker-desktop.enable ||
+        config.virtualisation.podman.enable ||
+        (config.virtualisation.docker.enable &&
+        (lib.versionAtLeast config.virtualisation.docker.package.version "25")))
+      )
+      {
+
+        # Related issues:
+        #   - https://github.com/nix-community/NixOS-WSL/issues/433
+        #   - https://github.com/NVIDIA/nvidia-container-toolkit/issues/452
+        #
+        # By setting `useWindowsDriver` to true, the Nvidia libraries are properly
+        # mounted on the container from the host.
+        wsl.useWindowsDriver = lib.mkIf config.hardware.nvidia-container-toolkit.enable true;
+
+      };
+}


loicreynier:
I just tested this following https://github.com/NixOS/nixpkgs/pull/312253 and can confirm it works fine with config.hardware.nvidia-container-toolkit.enable.

However, I don't think this should be limited only to wsl.docker-desktop.enable, since Podman users, for example, may want this feature by default without having to install Docker on their system.

ereslibre:
Thanks a lot for checking @loicreynier! Good point, I have updated the PR to take into account `config.virtualisation.podman.enable` and `config.virtualisation.docker.enable` if the docker package supports CDI (Docker >= 25).

modules/default.nix
```diff
@@ -3,6 +3,7 @@
     ./build-tarball.nix
     ./docker-desktop.nix
     ./interop.nix
+    ./nvidia-container-toolkit.nix
```
SuperSandro2000, line 6:
```suggestion
    ./nvidia.nix
```
lets name this just nvidia, incase we collect more nvidia things

modules/nvidia-container-toolkit.nix
```diff
@@ -0,0 +1,21 @@
+{ config, lib, ... }:
+with builtins; with lib; {
```
SuperSandro2000, line 2:
```suggestion
{
```
we shouldn't add more of those

modules/nvidia-container-toolkit.nix
```diff
@@ -0,0 +1,21 @@
...
+
+  config =
+    mkIf (config.wsl.enable &&
+          (config.wsl.docker-desktop.enable ||
+           config.virtualisation.podman.enable ||
+           (config.virtualisation.docker.enable &&
+            (versionAtLeast config.virtualisation.docker.package.version "25")))) {
```
SuperSandro2000, line 9:
```suggestion
  config = mkIf (
    config.wsl.enable && (
      config.wsl.docker-desktop.enable ||
      config.virtualisation.podman.enable ||
      (config.virtualisation.docker.enable && (versionAtLeast config.virtualisation.docker.package.version "25"))
    )
  ) {
```
or something like that

modules/nvidia-container-toolkit.nix
```diff
@@ -0,0 +1,21 @@
...
+      # Related issues:
+      #   - https://github.com/nix-community/NixOS-WSL/issues/433
+      #   - https://github.com/NVIDIA/nvidia-container-toolkit/issues/452
+      #
+      # By setting `useWindowsDriver` to true, the Nvidia libraries are properly
+      # mounted on the container from the host.
+      wsl.useWindowsDriver = mkIf config.hardware.nvidia-container-toolkit.enable true;
+
+    };
+
```
SuperSandro2000, line 20:
```suggestion
      wsl.useWindowsDriver = mkIf config.hardware.nvidia-container-toolkit.enable true;
    };
```

loicreynier:
Looks good to me! Tested on a Podman setup and works as expected.

modules/nvidia-container-toolkit.nix
ereslibre: Updated with `nixpkgs-fmt`.

573:
Can this be force-pushed ?

qwqawawow:
Hi,the file **nvidia-container-toolkit.json**  partly depends on nixos's nvidia driver?
Here is 
``` json
      {
        "containerPath": "/usr/bin/nvidia-smi",
        "hostPath": "/nix/store/5s8mapl88avcqrggazg5kvm1y6b36a5p-nvidia-x11-555.58-6.6.36-bin/bin/nvidia-smi",
        "options": [
          "ro",
          "nosuid",
          "nodev",
          "bind"
        ]
      },
      {
        "containerPath": "/usr/local/nvidia/lib",
        "hostPath": "/nix/store/qb18b5vz3rd6jch2klqfkr9zv9pxnj06-nvidia-x11-555.58-6.6.36/lib",
        "options": [
          "ro",
          "nosuid",
          "nodev",
          "bind"
        ]
      },
      {
        "containerPath": "/usr/local/nvidia/lib64",
        "hostPath": "/nix/store/qb18b5vz3rd6jch2klqfkr9zv9pxnj06-nvidia-x11-555.58-6.6.36/lib",
        "options": [
          "ro",
          "nosuid",
          "nodev",
          "bind"
        ]
      }
    


           
```

ereslibre:
> Excuse me, should the file nvidia-container-toolkit.json partly depend on NixOS's nvidia driver?

Yes. This is expected.

modules/nvidia.nix
```diff
@@ -0,0 +1,24 @@
...
+      )
+      {
+
+        # Related issues:
+        #   - https://github.com/nix-community/NixOS-WSL/issues/433
+        #   - https://github.com/NVIDIA/nvidia-container-toolkit/issues/452
+        #
+        # By setting `useWindowsDriver` to true, the Nvidia libraries are properly
+        # mounted on the container from the host.
+        wsl.useWindowsDriver = lib.mkIf config.hardware.nvidia-container-toolkit.enable true;
```
nzbr, line 21:
```suggestion
        wsl.useWindowsDriver = lib.mkDefault config.hardware.nvidia-container-toolkit.enable;
```

nzbr (changes requested):

modules/nvidia.nix
SuperSandro2000: This would also turn it off per default. This is an often pattern that is undesired, not sure what the correct variant is here.

nzbr: Maybe `mkIf ... (mkDefault ...)` then?

PR #481, opened 2024-05-29
fix: prevent leaking home when building without sandbox

terlar:
Multiple times when I tried to get my WSL image build working on a self-hosted GitHub Action running in a rootless docker container. I ended up with the super annoying message and failure:
```
error: home directory '/homeless-shelter' exists; please remove it to assure purity of builds without sandboxing
```

It was quite hard to pin-point where this came from, but I managed to reproduce it and saw that in this environment `/homeless-shelter` was writable and thus this derivation ended up writing some of it's cargo cache to there. When I set the `HOME` variable like this, the issue disappears and I am finally able to build my configuration.

### DIFF
--- a/utils/default.nix
+++ b/utils/default.nix
@@ -6,6 +6,10 @@ rustPlatform.buildRustPackage {
   src = ./.;
   cargoLock.lockFile = ./Cargo.lock;
 
+  preBuild = ''
+    export HOME=$(mktemp -d)
+  '';
+
   env = {
     NIXOS_WSL_SH = "${bash}/bin/sh";
     NIXOS_WSL_ENV = "${coreutils}/bin/env";


SuperSandro2000:
I think you need to solve this at your runner level otherwise many other packages will also encounter this issue.

terlar:
I haven't run into this for any other packages, so I don't think it is that common with things that tries to write to the unwrite-able home. But I guess in the normal/strict sandboxed environment it will prevent the writing without failure and in the unsandboxed environment it will just be allowed to write wherever.

I do remember having issues like this when I had unsandboxed environments that I added sandbox to and several tools failed because they tried to write to HOME.

I guess the cargo is lenient and don't blow up when trying to write to unwritable location.

SuperSandro2000:
> I guess the cargo is lenient and don't blow up when trying to write to unwritable location.

Yeah, it probably falls back to another path or doesn't write the cache at all.

utils/default.nix
```diff
@@ -6,6 +6,10 @@ rustPlatform.buildRustPackage {
   src = ./.;
   cargoLock.lockFile = ./Cargo.lock;
 
+  preBuild = ''
+    export HOME=$(mktemp -d)
```
nzbr, line 10:
```suggestion
    export HOME=${NIX_BUILD_TOP}/.home
    mkdir -p $HOME
```

It's probably better to put the HOME directory into the build path so that it may be cleaned up by nix with the rest of the build files. At least that's what I've always done in similar situations

Issue #486, opened 2024-06-07
Activating native systemd support breaks shell


aikooo7:
## Bug description

When enabling native systemd support and rebuild, switch and restart system you will be unable to log in to your system unless you use the recovery shell or use the shell binary directly instead of the wrapper.

Note: While trying to find what was the cause of the issue I rebuilt my home-manager and got ``Failed to connect to bus: No such file or directory`` so maybe #375 is related.

## To Reproduce
Add ``wsl.nativeSystemd = true;``
Run ``sudo nixos-rebuild switch``
Run ``wsl --shutdown`` in cmd.
Run ``wsl -d NixOS`` in cmd. NixOS should be your nixos distro's name in wsl.

## Logs
```
thread 'main' panicked at src/shell_wrapper.rs:105:10:
called `Result::unwrap()` on an `Err` value: When initializing journal logger

Caused by:
    No such file or directory (os error 2)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

## WSL version

2.0.9.0
```
WSL version: 2.0.9.0
Kernel version: 5.15.133.1-1
WSLg version: 1.0.59
MSRDC version: 1.2.4677
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.25131.1002-220531-1700.rs-onecore-base2-hyp
Windows version: 10.0.19045.4170
```

nzbr:
It looks like systemd hasn't started properly. That should obviously not crash the shell though, I'll fix that

nzbr:
As for the underlying problem: Can you send me the output of `dmesg` executed in another distro right after this happens. To reduce the size of the output, it probably makes sense to shut down WSL with `wsl --shutdown` before that

polybluez:
Is this the same error?

```
C:\Users\Admin>wsl
when trying to exec the wrapped shell

Caused by:
    No such file or directory (os error 2)
```

polybluez:
There is a workaround: first fire up WSL with a recovery shell `wsl -d NixOS --system --user root -- /mnt/wslg/distro/bin/nixos-wsl-recovery` and then enter WSL as usual

nzbr:
[...]

PR #496, opened 2024-06-15
fix: make wrapper not crash if logger setup fails

nzbr:
See also #486


K900 (approved):
The change is OK, but I'm not sure how helpful it'll be if systemd has failed to start, because we'll probably be shelling into a completely broken system...

PR #510, opened 2024-07-07
fix: change `opengl` to `hardware`

Parsifa1:
due to [this](https://github.com/NixOS/nixpkgs/commit/98cef4c27326d0f9e521654441929c1c7c64f8e9)


Parsifa1:
It seems that there is already a similar PR, so it is closed

PR #521, opened 2024-07-26
Fix user units with systemd 256

K900:
Starting with systemd 256, `user@%i` BindsTo `user-runtime-dir@%i`, so we can't just have the latter be skipped and still get a user service manager instance. So, replace it with a script that basically does the exact same thing, but "succeeds" when the directory is there.


nzbr:
Fan you add a comment for the empty string in there? I recon it's there so the original value is cleared (if this works the same way systemd overrides do), but it is non-obvious

K900:
Sure, done.

K900:
nixpkgs-fmt pls

PR #522, opened 2024-07-26
usbip: fix default snippetIpAddress command

terlar:
This fixes some bugs that I discovered in the default command.

The nix string escaping was wrong so the `\` got lost and when I tested this the command had a trailing space that made it fail still.

Also this didn't work when I was using network-mode mirrored since it includes metric in the output:

```
$ ip route list
default via 192.168.1.1 dev eth0 proto kernel metric 35
192.168.1.0/24 dev eth0 proto kernel scope link metric 291
192.168.1.1 dev eth0 proto kernel scope link metric 35
```

This also adds a target so you can batch restart the services. This is handy when having multiple YubiKeys and or switching between VPN. Perhaps there could even be some auto-restart thing. Will have to think about if that is a hack or not.

### DIFF
--- a/modules/usbip.nix
+++ b/modules/usbip.nix
@@ -38,9 +38,14 @@ in
     services.udev.enable = true;
 
     systemd = {
+      targets.usbip = {
+        description = "USBIP";
+      };
+
       services."usbip-auto-attach@" = {
         description = "Auto attach device having busid %i with usbip";
         after = [ "network.target" ];
+        partOf = [ "usbip.target" ];
 
         scriptArgs = "%i";
         path = with pkgs; [
--- a/modules/usbip.nix
+++ b/modules/usbip.nix
@@ -21,7 +21,7 @@ in
 
     snippetIpAddress = lib.mkOption {
       type = lib.types.str;
-      default = "$(ip route list | sed -nE 's/(default)? via (.*) dev eth0 proto kernel/\2/p')";
+      default = "$(ip route list | sed -nE 's/(default)? via (.*) dev eth0 .*/\\2/p' | head -n1)";
       example = "127.0.0.1";
       description = ''
         This snippet is used to obtain the address of the Windows host where Usbipd is running.


terlar:
Seems it is still brittle when connected to VPN with multiple routes. This solves it for me:
```nix
"$(ip route list | sed -nE 's/(default)? via (.*) dev .*/\\2/p' | head -n1)";
```

terlar:
Networking mode NAT, no VPN
```
$ wslinfo --networking-mode
nat
$ ip route list
default via 172.17.176.1 dev eth0 proto kernel
172.17.176.0/20 dev eth0 proto kernel scope link src 172.17.181.191
$ ip route list | sed -nE 's/(default)? via (.*) dev .*/\2/p' | head -n1
172.17.176.1
```

Networking mode mirrored, no VPN
```
...
```

Networking mode mirrored, VPN
```
...
```

terlar:
Is there any reason not to always grab the default route?

modules/usbip.nix
```diff
@@ -21,7 +21,7 @@ in
 
     snippetIpAddress = lib.mkOption {
       type = lib.types.str;
-      default = "$(ip route list | sed -nE 's/(default)? via (.*) dev eth0 proto kernel/\2/p')";
+      default = "$(ip route list | sed -nE 's/(default)? via (.*) dev .*/\\2/p' | head -n1)";
```
SuperSandro2000, line 24: I think we want to keep the eth0 and so on and just add the wildcard after it

terlar: What about the `head -n1`, otherwise you risk getting multiple matches, which won't work

terlar: Any reason to keep it at eth0 by the way? If not wsl-vpnkit would work out of the box.

terlar: @SuperSandro2000 Pinging in case you didn't see my follow-up questions.

SuperSandro2000: > What about the `head -n1`, otherwise you risk getting multiple matches, which won't work

yeah, lets add that.

> Any reason to keep it at eth0 by the way? If not wsl-vpnkit would work out of the box.

Thats the default interface WSL uses.

terlar: @SuperSandro2000 I adjusted it per your request. I am still not sure what the drawback is to not specify eth0 as it will work in all the cases. Instead of when you do specify it, those users won't have it working out of the box, but will have to adjust this configuration (which is not a big deal, but still).

SuperSandro2000 (approved):

PR #524, opened 2024-07-31
Load kernel modules

terlar:
WSL now supports kernel modules and come with a bunch configured as `m` and can be loaded this way. For example `vhci-hcd` used by usbip has changed from `y` to `m` and hence need to be explicitly loaded.

https://github.com/dorssel/usbipd-win/issues/995
https://github.com/microsoft/WSL/discussions/11789#discussioncomment-10083621

Unfortunately we cannot use `boot.kernelModules` since it is gated behind `boot.kernel.enable`.

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -10,6 +10,14 @@ in
   options.wsl = with types; {
     enable = mkEnableOption "support for running NixOS as a WSL distribution";
     useWindowsDriver = mkEnableOption "OpenGL driver from the Windows host";
+    kernelModules = mkOption {
+      type = listOf str;
+      default = [ ];
+      description = ''
+        The set of kernel modules to be loaded in the second stage of
+        the boot process via systemd-modules-load.service.
+      '';
+    };
     binShPkg = mkOption {
       type = package;
       internal = true;
@@ -114,6 +122,9 @@ in
         })
         (mkIf config.wsl.wslConf.network.generateResolvConf {
           "resolv.conf".enable = false;
+        })
+        (mkIf (cfg.kernelModules != [ ]) {
+          "modules-load.d/nixos.conf".text = concatStringsSep "\n" cfg.kernelModules;
         })
       ];
     };
--- a/modules/usbip.nix
+++ b/modules/usbip.nix
@@ -78,5 +78,7 @@ in
 
       targets.multi-user.wants = map (busid: "usbip-auto-attach@${busid}.service") cfg.autoAttach;
     };
+
+    wsl.kernelModules = [ "vhci-hcd" ];
   };
 }


modules/wsl-distro.nix
```diff
@@ -10,6 +10,14 @@ in
...
     enable = mkEnableOption "support for running NixOS as a WSL distribution";
     useWindowsDriver = mkEnableOption "OpenGL driver from the Windows host";
+    kernelModules = mkOption {
+      type = listOf str;
+      default = [ ];
+      description = ''
+        The set of kernel modules to be loaded in the second stage of
+        the boot process via systemd-modules-load.service.
+      '';
+    };
```
K900, line 20: Why can't we reuse `boot.kernelModules` here?

terlar:
Since it is gated behind `boot.kernel.enable`.

Another option is to have something like a boolean `wsl.loadKernelModules`. That will write the file with the value based on `boot.kernelModules`. What do you prefer?

modules/wsl-distro.nix
terlar: Since it is gated behind boot.kernel.enable which we don't want to enable.

Another option is to have something like a boolean wsl.loadKernelModules. That will write the file with the value based on boot.kernelModules. What do you prefer?

K900:
Wait, how does this even work. Isn't our modprobe looking in the wrong path for this?

nzbr:
Just wanted to ask this, because if I run `insmod vhci-hcd` it does not find the module file

K900:
You want `modprobe`.

terlar:
Systemd will do the loading based on config files located in this directory `/etc/modules-load.d`.

https://www.freedesktop.org/software/systemd/man/latest/modules-load.d.html

For NixOS it works like this:
- https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/boot/kernel.nix#L11
- https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/boot/kernel.nix#L321
- https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/boot/kernel.nix#L370

K900:
But how does it know where to find the WSL modules

terlar:
It is bundled with the kernel as loadable modules as I understand it and this change happened in the latest WSL kernel release (6.6)

nzbr:
> You want `modprobe`.

It seems that modprobe does not report if it failed though. It'll just run through but then `lsmod` will still show that no modules are loaded

K900:
WSL also rolled back to 5.15 for now. I really want to see a test at this point to actually verify it does what it says.

nzbr:
Ah, i have 5.15, that might be why this doesn't work

terlar:
Yes, that is a bit confusing, just discovered they reverted to the older kernel again now when I was looking into it. I have tested this on WSL with the 6.6 kernel and verified it works. But I guess we might as well wait looking into it since the latest WSL went back to the older kernel.

Stinjul:
WSL 2.5.7 switched to the 6.6 kernel, can this PR get merged now?

inschrift-spruch-raum:
The kernel upgrade seems to have stabilized. Should we proceed with it?

terlar:
I rebased it to make sure it is ready with the latest things.

PR #525, opened 2024-08-02
Support envfs

Atry:
Disable `populateBin` when `envfs` is enabled

### DIFF
--- a/modules/wsl-distro.nix
+++ b/modules/wsl-distro.nix
@@ -207,13 +207,24 @@ in
 
     # require people to use lib.mkForce to make it harder to brick their installation
     wsl = {
-      populateBin = true;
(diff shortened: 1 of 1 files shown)


modules/wsl-distro.nix
```diff
@@ -207,14 +207,25 @@ in
 
     # require people to use lib.mkForce to make it harder to brick their installation
     wsl = {
-      populateBin = true;
+      populateBin = !config.services.envfs.enable;
```
SuperSandro2000, line 210: This should use lib.mkIf and set the ture Boolean literal to prevent any unforseen condition fuckups.

Atry: Updated!

PR #537, opened 2024-09-02
shim: make systemd log to kmsg

### LINKED ISSUE #529: Make systemd log to kmsg
The option to make systemd log to kmsg should be passed through the shim, so that it is possible to debug it using `dmesg` in another distro

nzbr:
This makes it so that the systemd logs will be availably by running `dmesg` in another distro or the system distro

Closes #529


K900:
Oh god. I spent like two hours figuring out how to do that. And they just. Added a flag.

PR #538, opened 2024-09-06
Add troubleshooting tip for if systemd doesn't start with WSL

RossSmyth:
I was having this problem, and this was suggest to me and it worked. 
https://matrix.to/#/!apXxbtexmNGJaAryzM:nixos.org/$MR4egL2Tz6dIcnSLk-zRY14fRoj1NN2-7qc5sgJiW9A?via=nixos.org&via=matrix.org&via=beeper.com

I don't know the details of why it works, but it would be good to have this written down.

### DIFF
--- a/docs/src/troubleshooting/README.md
+++ b/docs/src/troubleshooting/README.md
@@ -12,3 +12,11 @@
   - To check which version you currently have installed, run `wsl --version`
     - The latest version can be found on the [Microsoft/WSL](https://github.com/microsoft/WSL/releases/latest) repo
     - If this command does not work, you are probably not using the Microsoft Store version of WSL!
+- If after booting into WSL and attempting to use the Nix daemon results in the output  
+  `error: cannot connect to socket at '/nix/var/nix/daemon-socket/socket': Connection refused`,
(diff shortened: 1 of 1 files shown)


RossSmyth:
@K900 might know?

K900:
Yes, systemd 256 requires cgv2. Also, I tried making a dynamic warning for this, but could not find any way to log things to the console during early boot :(

antoineco:
@K900 thanks! 

For whoever is interested in knowing the background for this change, I stumbled upon this: 
- https://github.com/systemd/systemd/issues/30852

(Linked from Arch's [wiki](https://wiki.archlinux.org/title/Cgroups#Enable_cgroup_v1))

If a warning was to be added to the README, it would be useful to mention:

- the condition in which the problem occurs (systemd 256, nixpkgs-unstable)
- the reason for the occurrence (WSL2 has cgroup v1 enabled by default)

nzbr:
Giving a reason in the docs and maybe a link here or to the upstream issue would be nice I think. Also this should maybe have its own heading instead of being mixed in with the general troubleshooting tips (it can stay on the same page though)

PR #560, opened 2024-10-04
wslConf: convert to freeform type

### LINKED ISSUE #559: Provide a `options.wsl.conf.extraOptions` value for other `/etc/wsl.conf` values
As per [microsoft documentation](https://learn.microsoft.com/en-us/windows/wsl/wsl-config), there are a lot of WSL configuration knobs you can set. If you use `config.wsl.enable` however, `/etc/wsl.conf` becomes a read-only symlink into the store, and there is no module option available to add your own customized options. This is a fairly simply problem to fix, and would greatly reduce maintainer burden on needing to keep up and update the module as microsoft changes things.

nzbr:
Closes #559

Pretty much what it says on the tin

### DIFF
--- a/modules/wsl-conf.nix
+++ b/modules/wsl-conf.nix
@@ -1,4 +1,4 @@
-{ lib, config, ... }:
+{ pkgs, lib, config, ... }:
 
 with lib; {
   imports = [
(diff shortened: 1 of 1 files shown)


K900:
Maybe make it proper RFC42 then?

nzbr:
Should've read that earlier 😅
Naming-wise: Should the options be moved to `wsl.settings` or stay as `wsl.wslConf` then? While the first one would be more in the spirit of the RFC from what i gathered so far, it doesn't make it as clear that the values end up in /etc/wsl.conf

K900:
Maybe `wsl.wslConf.settings`?

nzbr:
Wouldn't that be redundant? All the options currently in wsl.wslConf are 1:1 representations of the config file entries anyway, just with checks attached. I would keep the checks in place and just convert the whole thing to a freeform type

K900:
I guess we could just keep `wsl.wslConf` then.

nzbr:
Done, seems to work just fine from my quick test

SuperSandro2000 (approved):

PR #600, opened 2024-12-06
Distribute as .wsl file instead of .tar.gz

nzbr:
Basically this is just a renamed version of the existing `.tar.gz` with a manifest file bundled in, allowing for a more convenient installation.
See https://learn.microsoft.com/en-us/windows/wsl/build-custom-distro

### DIFF
--- a/.github/workflows/on_release.yml
+++ b/.github/workflows/on_release.yml
@@ -42,7 +42,7 @@ jobs:
 
       - name: Generate Checksums 🔑
         run: |
-          for x in *.{tar.gz,zip}; do
+          for x in *.{wsl,zip}; do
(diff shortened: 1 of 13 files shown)


nzbr:
Really hope I did not miss a `.tar.gz` somewhere

K900:
Does this still work on older WSL versions?

nzbr:
It should. The file is just a renamed tar.gz after all and wsl.exe does not seem to check the extension. The manifest will just be ignored if this is installed with `wsl --import`. The CI doesn't have the prerelease version with `--install --from-file` and it does work there

K900 (approved):
LGTM then.

PR #602, opened 2024-12-06
fix: shell-wrapper sometimes can't launch wrapped shell due to race condition

nzbr:
Also added early-boot (pre journald) logging for commands that are started as root


utils/src/shell_wrapper.rs
```diff
@@ -17,6 +20,13 @@ fn real_main() -> anyhow::Result<()> {
     // Therefore we dereference our symlink to get whatever it was originally.
     let shell = read_link(exe_dir.join("shell")).context("when locating the wrapped shell")?;
 
+    if shell.starts_with("/run/current-system/sw/bin/") {
+        while !Path::new("/run/current-system/sw/bin").exists() {
+            warn!("Activation script has not finished! Waiting for /run/current-system/sw/bin to exist");
```
K900, line 25: Can we do some inotify nonsense here? Or maybe write a oneshot pipe from the systemd wrapper?

utils/src/shell_wrapper.rs
```diff
@@ -100,16 +110,24 @@ fn real_main() -> anyhow::Result<()> {
...
-        .and_then(|logger| {
-            logger
-                .with_syslog_identifier("shell-wrapper".to_string())
-                .install()
-                .context("When installing journal logger")
-        })
-    {
-        warn!("Error while setting up journal logger: {:?}", err);
+    if nix::unistd::geteuid().is_root() {
+        // Journal isn't available during early boot
```
K900, line 114: Wouldn't this make all root shells log to kmsg? It's not awful, but suboptimal IMO. Maybe we should just use kmsg as a fallback for when the journal logger setup fails?

nzbr: It does, but kmsg also appears in the journal, so I didn't think about it too much. But it's probably better to just use kmsg as a fallback so we don't clutter kmsg even more. I'll change that

utils/src/shell_wrapper.rs
nzbr: inotify is probably easier than the pipe. I'll look into it

nzbr: The example code in the [nix crate's docs](https://docs.rs/nix/latest/nix/sys/inotify/index.html) looks like we'd still need a loop to check for inotify events, so I don't know how much would be gained by that. The filesystem call every few seconds shouldn't put the system under any kind of additional load and the race condition happens really infrequently anyway, at least in user facing shells (i've seen it in kmsg a few times)

K900: You do need a loop, but it will actually sleep exactly until the event happens, instead of artificially delaying for 3 seconds.

PR #613, opened 2025-01-05
refactor: drop flake-utils

isabelroses:
flake-utils' main utility is making flakes easier to make. but more modernly they don't provide much utility above of overrideable systems, which is not being utlised here since `eachSystem` is being used. So instead lets remove the input and vendor our own `forAllSystems` function to recreate what we previously used flake-utils for.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -16,24 +16,6 @@
         "type": "github"
       }
     },
-    "flake-utils": {
-      "inputs": {
-        "systems": "systems"
-      },
-      "locked": {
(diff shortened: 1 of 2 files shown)


nzbr:
Agreed, pulling in less dependencies is usually a good thing. I'm unsure if it wouldn't be better to keep the `forAllSystems` on the flake level like it was with flake-utils instead of running it once per output. Your version is definitely more readable, so I'd say we keep it like that unless it causes problems

PR #643, opened 2025-03-17
Add a logo

nzbr:
This has been lying around on my computer for quite some time, because I wasn't sure if I can use Segoe UI for this, but it seems to be OK according to [Microsoft's FAQ](https://learn.microsoft.com/en-us/typography/fonts/font-faq)


K900:
LGTM, love the logo

PR #665, opened 2025-06-09
fix: usbip is missing cat and ls

### LINKED ISSUE #662: Allow usbipd
## Problem
```
C:\Windows\System32>usbipd attach --wsl --busid 1-2
usbipd: info: Using WSL distribution 'NixOS' to attach; the device will be available in all WSL 2 distributions.
usbipd: error: WSL kernel is not USBIP capable; update with 'wsl --update'.

C:\Windows\System32>wsl --update
Checking for updates.
The most recent version of Windows Subsystem for Linux is already installed.

C:\Windows\System32>usbipd attach --wsl --busid 1-2
usbipd: info: Using WSL distribution 'NixOS' to attach; the device will be available in all WSL 2 distributions.
usbipd: error: WSL kernel is not USBIP capable; update with 'wsl --update'.

C:\Windows\System32>
```
## Solution

I need solution for redirect usb host to wsl nixos

nzbr:
Fixes #662 

Stacked on top of #666. That needs to be merged first

### DIFF
--- a/modules/usbip.nix
+++ b/modules/usbip.nix
@@ -45,9 +45,9 @@ in
     services.udev.enable = true;
 
     wsl.extraBin = [
-      { src = lib.getExe pkgs.coreutils; }
-      { src = lib.getExe pkgs.coreutils; }
-      { src = lib.getExe pkgs.kmod; }
+      { src = "${pkgs.coreutils}/bin/cat"; }
+      { src = "${pkgs.coreutils}/bin/ls"; }
+      { src = "${pkgs.kmod}/bin/modprobe"; }
     ];
 
     systemd = {


modules/usbip.nix
```diff
@@ -44,6 +44,12 @@ in
 
     services.udev.enable = true;
 
+    wsl.extraBin = [
```
SuperSandro2000, line 47: nit: we could use lib.getExe

SuperSandro2000 (approved):

modules/usbip.nix
nzbr: I always forget that that's an option

modules/usbip.nix
```diff
@@ -44,6 +44,12 @@ in
...
+      { src = lib.getExe pkgs.coreutils; }
```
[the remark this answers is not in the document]
nzbr: It also made no sense, because none of the three packages has a mainProgram set, so getExe doesn't work anyway. I just reverted the commit

PR #666, opened 2025-06-09
merge /bin and /sbin

nzbr:
This will allow us to make files available in `/sbin` using `extraBin`. usbip needs `/sbin/modprobe` to load `vhci_hcd` on demand

### DIFF
--- a/modules/systemd/native/default.nix
+++ b/modules/systemd/native/default.nix
@@ -24,6 +24,9 @@ with lib; {
           user.default = config.users.users.${cfg.defaultUser}.name;
           boot.systemd = true;
         };
+        extraBin = [
+          { src = "${pkgs.shadow}/bin/login"; }
+        ];
       };
 
       system.activationScripts = {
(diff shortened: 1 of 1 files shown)


modules/systemd/native/default.nix
```diff
@@ -24,23 +24,37 @@ with lib; {
...
+            if [ ! -d /sbin ]; then
+              rm -rf /sbin
+              mkdir -p /sbin
+            fi
+          ''
+        );
+        shimSystemd = stringAfter [ "createSbin" ] ''
           echo "setting up /sbin/init shim..."
-          mkdir -p /sbin
           ln -sf ${config.system.build.nativeUtils}/bin/systemd-shim /sbin/init
```
SuperSandro2000, line 56: And that will place things in /bin then?

SuperSandro2000 (approved):

modules/systemd/native/default.nix
nzbr: Yes, because this activation script runs after the one that creates the symling

K900:
Now that WSL is open source, should we maybe fix it upstream instead? https://github.com/microsoft/WSL/blob/ae323a4e9b0c1082a1bf1e050df4f88affa9f931/src/linux/init/main.cpp#L87

nzbr:
I thought this was because usbipd-win hard-coded that path, but I just re-checked and that doesn't seem to be the case:
https://github.com/dorssel/usbipd-win/blob/b13e3f3799e368835913e0468ded52fe1fbb97d2/Usbipd/Wsl.cs#L356
But I'm still not sure if changing the path in WSL would actually fix this -- it's been a while since I debugged it

Still I think just merging the two directories would be a good idea. It saves us from creating activation scripts for everything we need to have in /sbin in the future, because we can just use the existing extraBin option

K900:
I'd rather get rid of /sbin entirely.

nzbr:
Fair, but then we'd need another way of telling WSL where to look for the init binary. I don't see a good other way to do that. It might be possible through the config file, but not without breaking backwards compatibility for everyone who is still on an older WSL version. I'd be fine with that were it just for something like usbip or docker desktop support, but not with something that's an essential part of the boot code

K900:
AFAIUI the open source WSL can install on any supported version no?

SuperSandro2000:
> Now that WSL is open source, should we maybe fix it upstream instead? [microsoft/WSL@`ae323a4`/src/linux/init/main.cpp#L87](https://github.com/microsoft/WSL/blob/ae323a4e9b0c1082a1bf1e050df4f88affa9f931/src/linux/init/main.cpp#L87)

We cannot just change that as other distros have the binary there and relying on PATH maybe opens the next can of worms https://packages.debian.org/search?searchon=contents&keywords=modprobe&mode=path&suite=stable&arch=any

SuperSandro2000 (approved):

PR #752, opened 2025-08-21
feat: add --extra-files and --chown parameters

### LINKED ISSUE #612: Allow extra files in the tarball
## Problem
I would like to include extra files in the tarball without exposing them to the `/nix/store` (e.g. not using `environment.etc` or `systemd.tmpfiles`). 

Use case: include custom SSH host keys.

## Solution

I think the easiest thing would be to add a `--extra-files` flag to the `tarballBuilder` script, which is expected to be a folder with the correct directory structure inside, and then do `cp -R "${extraFiles}/." "$root"`. Similar to how it's already done for the config: https://github.com/nix-community/NixOS-WSL/blob/dee4425dcee3149475ead0cb6a616b8a028c5888/modules/build-tarball.nix#L87-L89

Inspired by NixOS-anywhere's `--extra-files`: https://github.com/nix-community/nixos-anywhere/blob/97b45ac774699b1cfd267e98a8bdecb74bace593/docs/howtos/extra-files.md

karaolidis:
Resolves #612.

### DIFF
--- a/docs/src/building.md
+++ b/docs/src/building.md
@@ -2,8 +2,7 @@
 
 This requires access to a system that already has Nix installed. Please refer to the [Nix installation guide](https://nixos.org/guides/install-nix.html) if that\'s not the case.
 
-If you have a flakes-enabled Nix, you can use the following command to
-build your own tarball instead of relying on a prebuilt one:
+If you have a flakes-enabled Nix, you can use the following command to build your own tarball instead of relying on a prebuilt one:
 
 ```sh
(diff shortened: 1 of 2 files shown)


nzbr:
Hey
Thanks for the PR and sorry for the late response
Code looks good, I just added a test to ensure we don't break the new options in the future and I fixed a small error in the `nix run` command in the docs.
By the way, thanks a lot for updating the docs as well!

karaolidis:
No problem, thank _you_ for the great work on this module! <3

PR #830, opened 2025-10-16
docs: add hint to set password when using sudo with password to docs

### LINKED ISSUE #829: Per default, the nixos user does not have a password. You should be able to use sudo without one. You may change the password with `sudo passwd nixos`, after that you can set `security.sudo.wheelNeedsPassword = true` if you want to be prompted for it when using sudo.
Per default, the nixos user does not have a password. You should be able to use sudo without one. You may change the password with `sudo passwd nixos`, after that you can set `security.sudo.wheelNeedsPassword = true` if you want to be prompted for it when using sudo.

I hope, that this answered your question. If not, [please provide additional details about your problem](https://stackoverflow.com/help/how-to-ask)

_Originally posted by @nzbr in https://github.com/nix-community/NixOS-WSL/discussions/71#discussioncomment-2501050_

This should go to https://github.com/nix-community/NixOS-WSL/blob/main/docs/src/install.md directly as a precaution. Otherwise a user might after initial install deploy a configuration.nix without a password set and at the same time `security.sudo.wheelNeedsPassword = true;` set.
Guess who did exactly that :D

EDIT: Workaround `wsl -d NixOS-WSL -u root` and therein `passwd nixos`

573:
Add note to set password before deploying new configuration, closes #829

### DIFF
--- a/docs/src/install.md
+++ b/docs/src/install.md
@@ -52,12 +52,6 @@ If you chose a different name for your distro during import, adjust this command
 
 ## Post-Install
 
-After the initial installation, you need to update your channels once, to be able to use `nixos-rebuild`:
-
-```sh
-sudo nix-channel --update
-```
-
 Don't forget to set a password for your user (default is `nixos` and is in the `wheel` group) in case you set `security.sudo.wheelNeedsPassword = true;` which is the default.
 
 You can set your password by running:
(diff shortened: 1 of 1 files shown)


docs/src/install.md
```diff
@@ -58,6 +58,8 @@ After the initial installation, you need to update your channels once, to be abl
 sudo nix-channel --update
 ```
 
+Don't forget to set a password for your user (default is `nixos` and is in the `wheel` group) in case you'll set ` security.sudo.wheelNeedsPassword = true;`.
```
SuperSandro2000:
```suggestion
Don't forget to set a password for your user (default is `nixos` and is in the `wheel` group) in case you'll set `security.sudo.wheelNeedsPassword = true;` which is the default.
```

PR #970, opened 2026-02-01
feat: wl-clipboard-wsl

### LINKED ISSUE #700: windows compatible wl-copy
Hi 

So I've made a custom `wl-clipboard` package which uses `clip.exe` and `powershell.exe -command Get-Clipboard` and I waned to ask if that'd be something that we could add to to wsl configuration options.

Here is the package that I've written: https://codeberg.org/oricat/nix-workstation/src/branch/main/hosts/wsl/wl-clipboard-win.nix

karaolidis:
Resolves #700.

### DIFF
--- a/packages/wl-clipboard-wsl/default.nix
+++ b/packages/wl-clipboard-wsl/default.nix
@@ -1,0 +1,20 @@
+{ pkgs
+, c ? "/mnt/c"
+, ...
+}:
+let
+  wl-copy = pkgs.writeShellScriptBin "wl-copy" ''
+    printf '%s' "$(cat)" | ${pkgs.dos2unix}/bin/unix2dos | ${c}/windows/system32/clip.exe
+  '';
+
+  wl-paste = pkgs.writeShellScriptBin "wl-paste" ''
+    ${c}/windows/system32/windowspowershell/v1.0/powershell.exe -NoProfile -Command Get-Clipboard | ${pkgs.dos2unix}/bin/dos2unix
+  '';
+in
+pkgs.symlinkJoin {
+  name = "wl-clipboard-wsl";
+  paths = [
+    wl-copy
+    wl-paste
+  ];
+}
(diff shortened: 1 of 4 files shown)


SymphonySimper:
Hello! could you replace `powershell.exe` and `clip.exe` with their absolute paths? The current implementation won't work if [wsl.wslConf.interop.appendWindowsPath](https://nix-community.github.io/NixOS-WSL/options.html#wslwslconfinteropappendwindowspath) is set to false. I have personally disabled this in my machine, as it slows down the PATH lookups.

- `powershell.exe` -> `/mnt/c/windows/system32/windowspowershell/v1.0/powershell.exe`
- `clip.exe` -> `/mnt/c/windows/system32/clip.exe`

packages/wl-clipboard-wsl/default.nix
```diff
@@ -0,0 +1,27 @@
...
+    wl-copy
+    wl-paste
+  ];
+
+  installPhase = ''
+    mkdir -p $out/bin
+    ln -s ${wl-copy}/bin/wl-copy $out/bin/wl-copy
+    ln -s ${wl-paste}/bin/wl-paste $out/bin/wl-paste
+  '';
+}
```
ljxfstorm:
```suggestion
pkgs.symlinkJoin {
  name = "wl-clipboard-wsl";
  paths = [
    wl-copy
    wl-paste
  ];
}
```
I suggest using [`pkgs.symlinkJoin`](https://nixos.org/manual/nixpkgs/stable/#trivial-builder-symlinkJoin) instead of `pkgs.stdenv.mkDerivation`, since the scripts are already defined via `pkgs.writeShellScriptBin`.

nzbr:
I just tested on my machine and the regular `wl-clipboard` utilities do seem to just work, so I don't really see why a custom solution would be neccessary

karaolidis:
I just tested it as well and it surprisingly did work. It wasn't this way when I first opened the PR, but I guess there's been updates to WSL since then to improve clipboard sharing? In any case, I suppose this PR is no longer necessary.

inschrift-spruch-raum:
Do we need to close the original issue?

PR #1010, opened 2026-03-15
fix: proper check for setting /sbin when changing populateBin from true to false

### LINKED ISSUE #1009: Enabling `services.envfs.enable` bricks the distribution
## Bug description

If I set `services.envfs.enable = true;` in my configuration after the switch the distribution won't load and I'll see only hanging screen of wsl. Any wsl command it attempt to load the distribution will hang, shutdown also hangs for a long time.

I was only able to restore the distribution by entering the vhd from another working distro and mapping the nix profile system to the old generation without envfs enabled. Note, all the previous generation that were worked before with envfs enabled became broken and only generation without envfs are working now.

## To Reproduce
I tested in details where the issue was introduced on the bare image of nixos-wsl.

1. Install 25.05 or later release
2. enable `services.envfs.enable` in the configuration.nix
3. rebuild and switch
4. system doesn't load

In comparison if I do the following:
1. Install 24.11 and older
2. enable `services.envfs.enable` in the configuration.nix
3. rebuild and switch
4. everything works with envfs
5. update to any later or unstable - everything works.

## Logs
I've attached my logs to another issue previously https://github.com/nix-community/NixOS-WSL/issues/966#issuecomment-4018641419 when I faced the issue, but I didn't found any issues or differences with the working distribution.

## WSL version

```
wsl --version
```

meskill:
fix #1009 

The bug 👿 introduced by https://github.com/nix-community/NixOS-WSL/pull/666/changes  

If populateBin is false (i.e. when services.envfs enabled for nixos) the check verifies if the `/sbin` is not folder, but if the system was built before without envfs the `/sbin` will be populated as symlink to `/bin`. 

After enabling envfs the check will fail because the symlink to dir is also a dir, so the sbin was always pointing to the empty bin managed by envfs. 

The fix verifies that the `/sbin` is recreated if it's a symlink or not directory. 

Verified on both latest bare nixos-wsl distro and my local setup that was previously bricking after every attempt to enable envfs.

### DIFF
--- a/modules/systemd/native/default.nix
+++ b/modules/systemd/native/default.nix
@@ -45,7 +45,7 @@ with lib; {
           ''
           else ''
             echo "setting up /sbin..."
-            if [ ! -d /sbin ]; then
+            if [ ! -d /sbin ] || [ -L /sbin ]; then
               rm -rf /sbin
               mkdir -p /sbin
             fi
--- a/tests/envfs/envfs.Tests.ps1
+++ b/tests/envfs/envfs.Tests.ps1
@@ -1,0 +1,37 @@
+BeforeAll {
+  . $PSScriptRoot/../lib/lib.ps1
+}
+
+Describe "envfs" {
+  BeforeAll {
+    $distro = [Distro]::new()
+  }
+
+  It "should install" {
+    $distro.InstallConfig("$PSScriptRoot/envfs.nix", "switch")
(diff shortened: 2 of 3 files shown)


nzbr:
Thanks!
I added a test for envfs so we don't break it again in the future and will merge this as soon as it has passed

PR #1013, opened 2026-03-16
Simplify implementation of interop.includePath

reasonableperson:
`modules/systemd/native/default.nix` currently prepends the string `$PATH` to the list of variables used by NixOS to set the PATH in `/etc/set-environment`. So when `/etc/set-environment` is sourced, the PATH set by WSL is "preserved" and prepended to the PATH generated by NixOS. The `split-path` utility is then used to rewrite the concatenated PATH, discarding paths beginning with `cfg.wslConf.automount.root` if `cfg.interop.includePath` is false. This has two odd side effects:

- Whether or not `cfg.interop.includePath` is set, the PATH will now include the [default path set by WSL](https://github.com/microsoft/WSL/blob/master/src/windows/service/exe/DistributionRegistration.h#L112), which is `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games`. These paths don't exist on NixOS.

- If a shell is started in a way that (unlike `wsl.exe`, hence #452) loads the [system-wide environment in `/etc/pam/environment`](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/config/system-environment.nix), such as `sudo su` or an ssh login, the PATH set by WSL will already have been replaced by NixOS by the time `/etc/set-environment` is sourced, and the hack in `modules/systemd/native/default.nix` will cause the PATH to be duplicated.

This change captures the old PATH in INTEROP_PATH instead, and [reverts](https://github.com/nix-community/NixOS-WSL/commit/deef7819e9863464365bc251c1070708e0925fc0) to a simple shell script to filter out the WSL defaults, removing `split-path` and its dependencies.


modules/interop.nix
```diff
@@ -32,6 +32,23 @@ with builtins; with lib;
         };
       };
 
+      environment = mkIf cfg.includePath {
+        # Capture Windows interop paths from the PATH set by WSL. Skip paths
+        # which don't start with /mnt/c; they are generic defaults (eg.
+        # /usr/local/sbin) that don't exist on NixOS.
+        variables.INTEROP_PATH = ''
+          $(echo -n "$PATH" \
```
K900, line 40: This is not going to work on non-bash shells.

K900: Even the thing we're currently doing is a horrible hack, frankly, and we should probably smuggle the right PATH value from systemd-shim or something.

modules/wsl-distro.nix
```diff
@@ -224,11 +224,22 @@ in
...
+    # WSL tools are symlinks to /init provided by WSL itself at /bin, but /bin
+    # is not in the NixOS PATH. Expose them via systemPackages so they land in
+    # /run/current-system/sw/bin instead.
+    environment.systemPackages = [
+      (pkgs.runCommand "wsl-binaries" { } ''
+        mkdir -p $out/bin
+        ln -s /init $out/bin/wslpath
+        ln -s /init $out/bin/wslinfo
+        ln -s /init $out/bin/mount.drvfs
+      '')
```
K900, line 236: I like this bit though.

PR #1029, opened 2026-04-19
docs(nix-flakes): say where the configuration file is

kant2002:
Specify the configuration file location for nixos-wsl. That's useful for complete novices which just get started, and maybe don't yet know where proper places to read about configuration located.

Other alternative is to provide link to official NixOS docs which explain how to edit configuration. But I didn't yet found this location

### DIFF
--- a/docs/src/how-to/nix-flakes.md
+++ b/docs/src/how-to/nix-flakes.md
@@ -1,6 +1,6 @@
 # How to configure NixOS-WSL with flakes
 
-First add a `nixos-wsl` input, then add `nixos-wsl.nixosModules.default` to your nixos configuration.
+First add a `nixos-wsl` input, then add `nixos-wsl.nixosModules.default` to your nixos configuration in `/etc/nixos/configuration.nix`.
 
 Below is a minimal `flake.nix` for you to get started:
 


docs/src/how-to/nix-flakes.md
```diff
@@ -1,6 +1,6 @@
 # How to configure NixOS-WSL with flakes
 
-First add a `nixos-wsl` input, then add `nixos-wsl.nixosModules.default` to your nixos configuration.
+First add a `nixos-wsl` input, then add `nixos-wsl.nixosModules.default` to your nixos configuration in `/etc/nixos/configuration.nix`.
```
lishaduck, line 3: This is for flakes, no? I'm moving from macOS to Windows and thus nix-darwin to NixOS-WSL, and this was very confusing. You can build from anywhere with flakes, although the wiki informs me that

> By default, [nixos-rebuild switch](https://wiki.nixos.org/wiki/Nixos-rebuild) will read its [configuration](https://wiki.nixos.org/wiki/Overview_of_the_NixOS_Linux_distribution#Declarative_Configuration_system_configuration) from /etc/nixos/flake.nix if it is present.

(TIL!)

PR #1044, opened 2026-04-28
fix(modules/wsl-distro): disable resolvconf if managed by WSL

K900:
nixpkgs added an assertion for this.


nzbr (approved):

FlorianFreudiger:
Problem: This also has the exact opposite effect:
If `wsl.wslConf.network.generateResolvConf = false;`
**then `networking.resolvconf.enable` is set to true instead of it's default value**.

Which in turn breaks my config since I use systemd-resolved, not resolvconf to manage my `/etc/resolv.conf`.

PR #1051, opened 2026-05-25
fix: condition channel addition on config

### LINKED ISSUE #1049: Channel copy behaviour hardcoded
## Problem

Channels are unconditionally included, even if entirely disabled on the host and the target. This leads to a non-blocking error when building.
```
[NixOS-WSL] Adding channel...
Failed to create stream fd: No such file or directory
```

## Solution

The channel copy logic at [`modules/build-tarball.nix#L158-L159`](https://github.com/nix-community/NixOS-WSL/blob/5482f113fd31ebac131d1ebeb2ae90bf0d5e41f5/modules/build-tarball.nix#L158-L159) should be gated behind a `config.nix.channel.enable` check.

----

I've tested this at [`x13-me/NixOS-WSL`](https://github.com/x13-me/NixOS-WSL/blob/00fe7ce2a8cbf98fcbcfb255c2dfeb4239a03012/modules/build-tarball.nix#L162-L166).

x13-me:
Resolves #1049 by gating the channel copy behind a check.

This is my first PR, and I open it as a draft for feedback

### DIFF
--- a/modules/build-tarball.nix
+++ b/modules/build-tarball.nix
@@ -155,8 +155,10 @@ in
           --system ${config.system.build.toplevel} \
           --substituters ""
 
-        echo "[NixOS-WSL] Adding channel..."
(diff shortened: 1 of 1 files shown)


SuperSandro2000 (approved):
LGTM

PR #1078, opened 2026-06-23
feat(wsl-conf): add boot.initTimeout option

karaolidis:
Adds a `wsl.wslConf.boot.initTimeout` option, mapping to the `[boot] initTimeout` key in `wsl.conf`.

WSL parses this key in `src/linux/init/WslDistributionConfig.cpp` (`ConfigKey("boot.initTimeout", BootInitTimeout)`, default `10 * 1000` ms), but is undocumented.

It bounds how long WSL waits for init to become ready at boot and, since microsoft/WSL#13552, how long `systemctl poweroff` is given to run the shutdown sequence before WSL forces `reboot(RB_POWER_OFF)` (see also microsoft/WSL#13493).

Raising it lets services with longer `ExecStop` handlers shut down cleanly.

### DIFF
--- a/modules/wsl-conf.nix
+++ b/modules/wsl-conf.nix
@@ -63,6 +63,16 @@ with lib; {
               default = true;
               description = "Use systemd as init. Disabling this may break your NixOS installation";
             };
+            initTimeout = mkOption {
+              type = types.nullOr types.ints.unsigned;
+              default = null;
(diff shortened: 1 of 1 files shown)


nzbr:
Oh neat! Good find!
This should be helpful against user-sessions not working if init takes too long

PR #1082, opened 2026-07-02
Add required binaries for Docker Desktop v4.80.0+ integration

### LINKED ISSUE #1081: Docker Desktop 4.80.0
I'm not 100% sure, this might have happened because of something else, but after updating to Docker Desktop 4.80.0 the integration stopped working. After a bit of debugging, it looks like it tries to use `install` and `mv` which aren't available on path the way it calls them. I was able to fix the issue by adding
```nix
wsl.extraBin = [
  { src = "${pkgs.coreutils}/bin/install"; }
  { src = "${pkgs.coreutils}/bin/mv"; }
];
```

tminich:
Fixes #1081


yswtrue:
It works for me.

SuperSandro2000 (approved):

